PR #30, opened 2014-08-29
fix getUncommittedEvents being used to also clear the uncomitted events

kelvinj:
added the following methods to `AggregateRoot` and therefore to `EventSourcedAggregateRoot`:
- `hasUncommittedEvents`
- `clearUncommittedEvents`


drgomesp:
What about something like `flushEvents` which could clear and return the uncommitted ones?

PR #51, opened 2014-09-16
Add doctrine readmodel support

Miliooo:
I'm already using this in my projects. (Although I'm considering using only dbal)
Before I write unit tests I want to make sure you guys want to add this to this library.

Todo
- [x] check if this is something broadway wants to support
- [ ] write tests
- [ ] make it configurable in the bundle


asm89:
Closing this for now. Feel free to re-open when you continue work on it. :)

I think it would be a nice addition.

PR #82, opened 2014-11-20
Make the simple command bus more resilient.

kimlai:
If an exception occurs during the handling of a command, the simple
command bus becomes unusable (`$isDispatching` stays `true`).
With this change the bus will be able to dispatch subsequent commands
even if an exception occurs.


macintoshplus:
@kimlai How to help you ?

I am greatly interested by this PR.

PR #102, opened 2014-12-04
Aggregate Root Test Scenario

MarkRedeman:
The current `EventSourcing/Testing/Scenario` seems to have fallen behind. This pull request added some functionality such that you can easily test your `EventSourcedAggregateRoot` objects without using a command bus.

Nb: I also added an example which was basically copied from the `event-sourced-domain-with-test` but changed such that it uses the new `AggregateRootScenarioTestCase` class.
It might be better to use a different example and also add a testcase for `EventSourcedEntity`.

### DIFF
--- a/src/Broadway/EventSourcing/Testing/AggregateRootScenarioTestCase.php
+++ b/src/Broadway/EventSourcing/Testing/AggregateRootScenarioTestCase.php
@@ -1,0 +1,59 @@
+<?php
+
+/*
+ * This file is part of the broadway/broadway package.
+ *
+ * (c) Qandidate.com <opensource@qandidate.com>
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+namespace Broadway\EventSourcing\Testing;
+
+use PHPUnit_Framework_TestCase as TestCase;
+use Broadway\EventSourcing\AggregateFactory\PublicConstructorAggregateFactory;
+
+/**
+ * Base test case that can be used to set up a command handler scenario.
+ */
+abstract class AggregateRootScenarioTestCase extends TestCase
+{
+    /**
+     * @var Scenario
+     */
+    protected $scenario;
+
+    public function setUp()
+    {
+        $this->scenario = $this->createScenario();
+    }
+
+    /**
+     * @return Scenario
+     */
+    protected function createScenario()
+    {
+        $aggregateRootClass = $this->getAggregateRootClass();
+        $factory            = $this->getAggregateRootFactory();
+
+        return new Scenario($this, $factory, $aggregateRootClass);
+    }
+
+    /**
+     * Returns a string representing the aggregate root
+     *
+     * @return string AggregateRoot
+     */
+    abstract protected function getAggregateRootClass();
+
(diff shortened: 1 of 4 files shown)


MarkRedeman:
I've updated the PR.
One thing I'm not really happy about is the code duplication in the `InviteTest` and `InvitationCommandHandlerTest` test. Both files test the same scenarios, but on a different level. 
It might be possible to move some of the `given()` and `then()` statements to a separate class, but it will lower the readability of the tests.

asm89:
@MarkRedeman thanks for the update! Don't worry about that duplication, it's example code and it's clearer this way. When you're testing an actual implementation you probably just use the version you're interested it.

Can you squash your commits? I'd be happy to get this merged.

asm89:
:+1: this is good to go.

ping @qandidate-labs/broadway

examples/event-sourced-domain-with-tests/InvitesTest.php
```diff
@@ -2,17 +2,16 @@
 
 require_once __DIR__ . '/Invites.php';
 
-/**
- * We drive the tests of our aggregate root through the command handler.
+/**.
  *
- * A command handler scenario consists of three steps:
+ * A aggreagte root scenario consists of three steps:
```
othillo: typo, sorry :)

examples/event-sourced-domain-with-tests/InvitesTest.php
```diff
@@ -22,11 +21,9 @@ public function setUp()
         $this->generator = new Broadway\UuidGenerator\Rfc4122\Version4Generator();
     }
 
-    protected function createCommandHandler(Broadway\EventStore\EventStoreInterface $eventStore, Broadway\EventHandling\EventBusInterface $eventBus)
+    protected function getAggegateRootClass()
```
othillo: getAggregateRootClass

src/Broadway/EventSourcing/Testing/AggregateRootScenarioTestCase.php
```diff
@@ -0,0 +1,59 @@
...
+
+        return new Scenario($this, $factory, $aggregateRootClass);
+    }
+
+    /**
+     * Returns a string representing the aggregate root
+     *
+     * @return string AggregateRoot
+     */
+    abstract protected function getAggegateRootClass();
```
othillo: getAggregateRootClass

src/Broadway/EventSourcing/Testing/AggregateRootScenarioTestCase.php
```diff
@@ -0,0 +1,59 @@
...
+    {
+        $this->scenario = $this->createScenario();
+    }
+
+    /**
+     * @return Scenario
+     */
+    protected function createScenario()
+    {
+        $aggregateRootClass = $this->getAggegateRootClass();
```
othillo: getAggregateRootClass

MarkRedeman:
@othillo thanks, just fixed them. :)

othillo:
:+1:

wjzijderveld:
Thanks @MarkRedeman!

Issue #109, opened 2014-12-20
Elastic search and findBy fields


harikt:
Hi, 

I was searching via `findBy` method https://github.com/qandidate-labs/broadway/blob/41a495d69fba2303622febfb3bc715a0f500c61d/src/Broadway/ReadModel/ElasticSearch/ElasticSearchRepository.php#L91 and was not getting any results. 

```<?php
$fields = [
    'name' => 'beta',
];
$this->repository->findBy($fields)
```

I tried with plain elastic search api and results are getting for below api .

```php
<?php
$searchParams = array();
$searchParams['index'] = 'something';
$searchParams['type']  = 'Some\Namespaced\Class';
$searchParams['body']['query']['wildcard']['name'] = 'b*';
// $searchParams['body']['query']['match']['name'] = 'beta';
```

The find with id is working as expected.

Thanks!

asm89:
@mbadolato When would these checks be run? Running them on each update operation would be quite expensive. We could add a default cli command to the bundle, but you'd still have to run it.

wjzijderveld:
Closing this for now, as the issue was in Elasticsearch mapping.
The issue we have in Broadway is lack of good documentation...

manelephant:
I can't solve this issue. Trying to set some field as "notAnalysedFields" using a class extending ElasticSearchRepository

 <service id="my_app.read_model.repository.card_overview"
                 class="MyApp\ReadModel\Repository\CardOverview">
        </service> 

Then using : 

        <service id="read_model.index_generator" class="MyApp\ReadModel\ElasticSearchIndexGenerator">
            <argument type="collection">
                <argument type="service" id="my_app.read_model.repository.card_overview" />
            </argument>
        </service>

But it keep crashing with this error : 

Catchable Fatal Error: Argument 1 passed to MyApp\ReadModel\Repository\CardOverview::__construct() must be an instance of Elasticsearch\Client,....etc

My class looks like this : 

class CardOverview extends ElasticSearchRepository
{
    public function __construct(
        Client $client,
        SerializerInterface $serializer,
        $index,
        $class,
        array $notAnalyzedFields = array('code')
    )
    {
        parent::__construct(
            $client,
            $serializer,
            $index,
            $class,
            $notAnalyzedFields
        );
    }
}

Is there something obvious for one of you guys ? Thanx :)

PR #110, opened 2014-12-24
Extract reusable ElasticSearch client handling from the implementation

simensen:
This could prove to be useful in the case that someone wants to leverage existing ElasticSearch code without extending the base ElasticSearch repository implementations. In particular, this is useful if someone does not have a need or desire to extend the base class but still wants to use some of its tools without having to copy & paste a LOT of code.

The name could possibly use some work. It might also want to live elsewhere. It is currently as tested as it was before, but only by way of being used in the same way. It might make sense to also extract the tests so that they are testing the newly extracted class rather than only testing it *through* the ElasticSearch repository implementation.

It might also make sense to create a factory to create instances of these. :)

refs #109

### DIFF
--- a/src/Broadway/ReadModel/ElasticSearch/ElasticSearchRepository.php
+++ b/src/Broadway/ReadModel/ElasticSearch/ElasticSearchRepository.php
@@ -22,29 +22,11 @@ use Elasticsearch\Common\Exceptions\Missing404Exception;
  */
 class ElasticSearchRepository implements RepositoryInterface
 {
-    private $client;
-    private $serializer;
-    private $index;
-    private $class;
-    private $notAnalyzedFields;
+    private $objectRepository;
 
-    /**
-     * @param string $index
-     * @param string $class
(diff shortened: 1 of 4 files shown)


wjzijderveld:
I kinda see why you want it within a larger application, but I'm not convinced it would benefit Broadway.
I don't really like the idea that Broadway should support all kinds of querying on all kinds of read model storage repositories.

asm89:
@wjzijderveld Maybe we should put something like that in the read model component docs and close this? What do you think @simensen?

simensen:
Yeah, I think that it should be made extremely clear to new people that the read model implementation is very basic and that if someone needs to extend it they are better of copying & pasting *the implementation* (and NOT the interfaces) and building from there. Just knowing that was the state of the world would have kept me from getting blocked on this at all. I would have known right away I should just roll w/ my own!

wjzijderveld:
Closing this for now

PR #73, opened 2014-10-31
[WIP] [RFC] EventStore management

kelvinj:
Allows events to be pulled out of the EventStore using a fluid criteria builder.

- includes a CallableEventVisitor, so event's can be easily processed using a callback function.
- no tests as of yet
- beyond the Axon implementation, added ability to query a property using `beginsWith`, `contains` and `endsWith`. In DBAL implementation, this correlates with `LIKE` statements.

Example of usage:

```php
<?php
$visitor = new CallableEventVisitor(function (DomainMessage $domainMessage) { /* do something */ });
$criteriaBuilder = $eventStore->newCriteriaBuilder();
$type = $criteriaBuilder->property('type'); // corresponds to column in events table
$criteria = $type->beginsWith('MyBoundedContext.AggregateRoot1.')->orWith(
    $type->beginsWith('MyBoundedContext.AggregateRoot2.')
);
$eventStore->visitEvents($visitor, $criteria);
```

*Heavily* based on AxonFramework's implementation.

*PS: having issues installing `instaclick/base-test-bundle` dependency at the moment, so no idea if this will interfere with tests passing, but at this point it's about getting feedback.*

PR #112, opened 2015-01-01
Create a method to retrieve the unique set of stream ids from the event store

mbadolato:
I created a helper method in the event store implementations to provide back the unique set of stream ids in the event store.

My use case for this is that I'm building a symfony command for rebuilding projections and I needed to be able to retrieve the stream id's so that I could then rebuild each event stream

```php
foreach ($eventStore->getStreamIds() as $id) {
    $eventStore->load($id);
    $eventBus->publish(...);
    // etc
}
```

### DIFF
--- a/src/Broadway/EventStore/DBALEventStore.php
+++ b/src/Broadway/EventStore/DBALEventStore.php
@@ -101,6 +101,22 @@ class DBALEventStore implements EventStoreInterface
         }
     }
 
+    /**
+     * {@inheritDoc}
+     */
+    public function getStreamIds()
+    {
+        $statement = $this->connection->prepare('SELECT DISTINCT uuid FROM ' . $this->tableName);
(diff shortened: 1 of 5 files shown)


asm89:
I'm not in favor of adding this. For the DBAL backed storage the implementation is quite straight forward, but for other implementations we might come up with one day probably not so much. Adding it to the `EventStoreInterface` is quite a big step.

The same result would be possible when https://github.com/qandidate-labs/broadway/pull/73 gets ready for prime time.

mbadolato:
As we (me, @asm89 , @wjzijderveld) were discussing in IRC, if this doesn't seem like something for Broadway and/or #73 will provide the same sort of functionality, I can go ahead and table this. I can decorate the the event store locally in my app to provide the functionality for myself for the time being :smile:  Closing for now.

PR #113, opened 2015-01-07
Child entity example.

simensen:
The domain model is clearly contrived for this example but I wanted to keep it simple. This is an example of a simple single child entity of an aggregate root.


asm89:
:+1: thank you for adding this!

wjzijderveld:
Thanks @simensen

Issue #61, opened 2014-09-25
Proper way to use the Metadata components?


mbadolato:
There are a couple of simple examples and tests where you guys are directly calling`DomainMessage::recordNow()` and passing in data to be used as Metadata, but there's no real world examples, and in fact, the task looks to be impossible in the course of the normal flow.  

Using `EventSourcedAggregateRoot` the only time `DomainMessage::recordNow()` gets triggered is when the AR methods of a class that extends `EventSourcedAggregateRoot` fire off a `apply($myEvent)` command.  But `apply()` does not accept a parameter for any metadata, and furthermore, hardcodes `new Metadata(array())` in the call to `recordNow`.

Is this intentional, or a bug/oversight?

We (or at least from what I understand to be the proper use of Broadway) are not making direct calls to `recordNow()` and should be using the library the way I described, via `apply()`. If that's the case, how do we apply metadata? 

To me, it would seem appropriate to change `apply()` to something like

``` php
public function apply($event, $metadata = [])
{
    $this->handleRecursively($event);

    $this->playhead++;
    $this->uncommittedEvents[] = DomainMessage::recordNow(
        $this->getAggregateRootId(),
        $this->playhead,
        new Metadata($metadata),
        $event
    );
}
```

Also, while on the subject, `EventSourcedAggregateRoot` implements `AggregateRoot`.  `AggregateRoot` does not define `apply()` as part of the interface. Shouldn't it? `EventSourcedAggregateRoot` is the only concrete implementation of `AggregateRoot` currently in the system, but it would seem to me that all AR's need a way to apply events to themselves.

Thoughts (on both issues)?  I'd be happy to submit a PR tonight.

PR #115, opened 2015-01-13
Publish decorated event stream

mbadolato:
Small change to have EventSourcingRepository publish the decorated event stream instead of the undecorated one.

The current implementation (non-decorated) doesn't allow for access to any metadata from the projectors on the initial publish. The metadata is available in the projectors on subsequent publishes (e.g., if the read model is rebuilt).

The decorated stream is already available in the repository at the time of publishing, however it was only being used in the ```append()``` operation and not the ```publish()``` that followed (the pre-decoration array of events was instead published).  So this adds no additional overhead; it merely makes more information available (information that is being saved to the database, and is available on subsequent publishes)

Original discussion of this missing functionality began [here](https://github.com/qandidate-labs/broadway/issues/61#issuecomment-57801381), and further research [here](https://github.com/qandidate-labs/broadway/issues/61#issuecomment-58119917) indicates that metadata should be available for use with projections

### DIFF
--- a/test/Broadway/EventSourcing/AbstractEventSourcingRepositoryTest.php
+++ b/test/Broadway/EventSourcing/AbstractEventSourcingRepositoryTest.php
@@ -18,15 +18,29 @@ use Broadway\Domain\DomainMessage;
 use Broadway\Domain\Metadata;
 use Broadway\EventHandling\SimpleEventBus;
 use Broadway\EventHandling\TraceableEventBus;
+use Broadway\EventSourcing\AggregateFactory\PublicConstructorAggregateFactory;
+use Broadway\EventSourcing\MetadataEnrichment\MetadataEnricherInterface;
+use Broadway\EventSourcing\MetadataEnrichment\MetadataEnrichingEventStreamDecorator;
 use Broadway\EventStore\EventStoreInterface;
 use Broadway\EventStore\InMemoryEventStore;
 use Broadway\EventStore\TraceableEventStore;
+use Broadway\ReadModel\Projector;
 use Broadway\TestCase;
+use RuntimeException;
 
 abstract class AbstractEventSourcingRepositoryTest extends TestCase
 {
+    /** @var TraceableEventBus */
+    protected $eventBus;
+
+    /** @var TraceableEventStoreDecorator */
+    protected $eventStreamDecorator;
+
     /** @var EventStoreInterface */
     protected $eventStore;
+
+    /** @var EventSourcingRepository */
+    protected $repository;
 
     public function setUp()
     {
@@ -44,7 +58,7 @@ abstract class AbstractEventSourcingRepositoryTest extends TestCase
 
     /**
      * @test
-     * @expectedException Assert\InvalidArgumentException
+     * @expectedException \Assert\InvalidArgumentException
      * @dataProvider objectsNotOfConfiguredClass
      */
     public function it_throws_an_exception_when_adding_an_aggregate_that_is_not_of_the_configured_class($aggregate)
@@ -96,7 +110,7 @@ abstract class AbstractEventSourcingRepositoryTest extends TestCase
 
     /**
      * @test
-     * @expectedException Broadway\Repository\AggregateNotFoundException
+     * @expectedException \Broadway\Repository\AggregateNotFoundException
      */
     public function it_throws_an_exception_if_aggregate_was_not_found()
     {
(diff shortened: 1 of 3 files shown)


asm89:
@mbadolato this makes total sense to me. Can you add a test verifying that it's the decorated event stream that is being published?

mbadolato:
@asm89 Sure, can look at it tonight

mbadolato:
Also, just a stylistic preference, but since ```$domainEventStream``` would only be used once after the change, do you mind if I remove it as an intermediate variable, or would you prefer to keep it?

From:

```php
$domainEventStream = $aggregate->getUncommittedEvents();
$eventStream       = $this->decorateForWrite($aggregate, $domainEventStream);
$this->eventStore->append($aggregate->getAggregateRootId(), $eventStream);
$this->eventBus->publish($eventStream);
```

To:
```php
$eventStream = $this->decorateForWrite($aggregate,  $aggregate->getUncommittedEvents());
$this->eventStore->append($aggregate->getAggregateRootId(), $eventStream);
$this->eventBus->publish($eventStream);
```

othillo:
@mbadolato removing the intermediate is fine.

asm89:
@mbadolato alternatively you rename the variables a bit and make the code more clear imo. :)

```
$eventStream = ...;
$decoratedEventStream = ...;
```

mbadolato:
@asm89 Either way works for me!

mbadolato:
@%@&%!&^$ php 5.3  and it's lack of short array syntax

cordoval:
what is @%@&%!&^$ ?

mbadolato:
| what is @%@&%!&^$ ?

A polite way of screaming a string of profanities in written form in a public medium

mbadolato:
@asm89 Ok, test written, commits squashed. Should be good to go.

test/Broadway/EventSourcing/AbstractEventSourcingRepositoryTest.php
```diff
@@ -25,6 +26,9 @@
 
 abstract class AbstractEventSourcingRepositoryTest extends TestCase
 {
+    /** @var EventBusInterface */
```
asm89: Can you change this to `TracableEventBus`? The method `->getEvents()` is called and only available on the `TraceableEventBus`. Feel free to change it for the other typehint as well.

mbadolato: Yep will do

test/Broadway/EventSourcing/EventSourcingRepositoryTest.php
```diff
@@ -78,6 +81,34 @@ public function it_throws_an_exception_if_the_static_method_does_not_exist()
         $repository->load('y0l0');
     }
 
+    /**
+     * @test
+     */
+    public function it_publishes_decorated_events()
```
asm89: Can you move this test to the base class? I think the base class exists for historical reasons, but I'd rather have the test there for now.

test/Broadway/EventSourcing/EventSourcingRepositoryTest.php
```diff
@@ -78,6 +81,34 @@ public function it_throws_an_exception_if_the_static_method_does_not_exist()
...
+    {
+        $projector = new TestMetadataPublishedProjector();
+        $this->eventBus->subscribe($projector);
+
+        $repository = new EventSourcingRepository(
+            $this->eventStore,
+            $this->eventBus,
+            '\Broadway\EventSourcing\TestEventSourcedAggregate',
+            new PublicConstructorAggregateFactory(),
+            array(new MetadataEnrichingEventStreamDecorator(array(new TestDecorationMetadataEnricher())))
```
asm89: If you move the test to the base class you can always construct the repository with the enricher.

test/Broadway/EventSourcing/EventSourcingRepositoryTest.php
```diff
@@ -78,6 +81,34 @@ public function it_throws_an_exception_if_the_static_method_does_not_exist()
...
+        );
+
+        $aggregate = $this->createAggregate();
+        $aggregate->apply(new DidNumberEvent(42));
+        $repository->add($aggregate);
+
+        /** @var Metadata $metadata */
```
asm89: We don't add these comments in the middle of code. Feel free to add a doc block to the projector (that should fix it as well?).

asm89:
@mbadolato See comments. Thank you for the added tests. :+1:

test/Broadway/EventSourcing/AbstractEventSourcingRepositoryTest.php
mbadolato: Done

test/Broadway/EventSourcing/EventSourcingRepositoryTest.php
mbadolato: Done

test/Broadway/EventSourcing/EventSourcingRepositoryTest.php
mbadolato: Can you explain further? The repository needs a decorator and the decorator needs an array of enrichers. I'm not sure what you meant by your comment, nor why moving it would change how it's constructed.

test/Broadway/EventSourcing/EventSourcingRepositoryTest.php
mbadolato: Done

mbadolato:
@asm89 Done!

othillo:
:+1:

wjzijderveld:
Thanks @mbadolato

mbadolato:
:smile:

PR #132, opened 2015-01-31
Getter fot the AggregateRootInstance in the testing scenario

bartdezwaan:
This might be handy if you want to check that your apply methods did there job.

### DIFF
--- a/src/Broadway/EventSourcing/Testing/Scenario.php
+++ b/src/Broadway/EventSourcing/Testing/Scenario.php
@@ -121,6 +121,14 @@ class Scenario
     }
 
     /**
+     * @return Broadway\EventSourcing\EventSourcedAggregateRoot
+     */
+    public function getAggregateRootInstance()
+    {
+        return $this->aggregateRootInstance;
+    }
+
+    /**
      * @return array Payloads of the recorded events
      */
     private function getEvents()


wjzijderveld:
What is your use-case exactly?
Wondering because the scenario is used to test emitted events after an action, testing if your aggregate is in the correct state based on given events should be a separate test imo.

bartdezwaan:
I do understand your thinking.
But the scenario tests are very descriptive. So I would prefer to use them even if I want to test the state of my aggregate. I see no reason why I wouldn't.

asm89:
@bartdezwaan In the event sourced case I don't see much need for it? The idea is that you assert the events emitted and the intermediate state shouldn't really matter? Do you have an example of where you think it does?

If you have an aggregate root that is _not_ event sourced I see that this could be useful, but in that case I'm not sure if this testing scenario is of any help.

bartdezwaan:
@asm89  I don't have a specific case.
But if I want to test my apply methods, I want to test them with an AR in different states.
Why not describe that states with the scenario functionality that is already there.

I just want my tests to read nicely.
If I am not able to use the scenario, I will have to build the AR myself or write a builder for it.

wjzijderveld:
Closing this, as we don't see the use of this in an event-sourced context.

PR #133, opened 2015-02-03
php-cs-fixer fixes

mbadolato:
Ran php-cs-fixer with the .php_cs file

### DIFF
--- a/examples/event-sourced-child-entity/Parts.php
+++ b/examples/event-sourced-child-entity/Parts.php
@@ -71,8 +71,8 @@ class Manufacturer extends Broadway\EventSourcing\EventSourcedEntity
 
     public function __construct($partId, $manufacturerId, $manufacturerName)
     {
-        $this->partId = $partId;
-        $this->manufacturerId = $manufacturerId;
+        $this->partId           = $partId;
+        $this->manufacturerId   = $manufacturerId;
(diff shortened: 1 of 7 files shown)


examples/event-sourced-child-entity/Parts.php
```diff
@@ -141,7 +141,7 @@ class PartManufacturerWasRenamedEvent
 
     public function __construct($partId, $manufacturerName)
     {
-        $this->partId = $partId;
+        $this->partId           = $partId;
```
cordoval: no please

mbadolato: That's part of the php_cs definition that Broadway uses. 

Some of us prefer having them lined up, some don't. But the broadway team chose to put that in their coding standards file. Personal preference is irrelevant.

cordoval: :baby: yes i remember now :+1: it is ok!

wjzijderveld:
Already noticed some of these at time of the PR, but didn't want to comment it at the time.
I'm not sure if we should be more strict during PR review.

I would be fine if we keep it as clean as possible, but don't comment everything on every PR. Just running PHPCS once in a while might be sufficient.

 Once we have a stable 1.0 that would probably change.

Thanks @mbadolato

PR #135, opened 2015-02-10
Make the simple event bus more resilient.

kimlai:
Extremely similar to https://github.com/qandidate-labs/broadway/pull/119.

The simple event bus should be able to continue publishing events even
if an exception occurs in an event listener.

### DIFF
--- a/test/Broadway/EventHandling/SimpleEventBusTest.php
+++ b/test/Broadway/EventHandling/SimpleEventBusTest.php
@@ -104,6 +104,41 @@ class SimpleEventBusTest extends TestCase
         $this->eventBus->publish($domainEventStream);
     }
 
+    /**
+     * @test
+     */
+    public function it_should_still_publish_events_after_exception()
+    {
+        $domainMessage1 = $this->createDomainMessage(array('foo' => 'bar'));
+        $domainMessage2 = $this->createDomainMessage(array('foo' => 'bas'));
+
+        $domainEventStream1 = new DomainEventStream(array($domainMessage1));
+        $domainEventStream2 = new DomainEventStream(array($domainMessage2));
+
+        $eventListener = $this->createEventListenerMock();
+        $eventListener
+            ->expects($this->at(0))
+            ->method('handle')
+            ->with($domainMessage1)
+            ->will($this->throwException(new \Exception('I failed.')));
+
+        $eventListener
+            ->expects($this->at(1))
+            ->method('handle')
+            ->with($domainMessage2);
+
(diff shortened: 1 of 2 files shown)


test/Broadway/EventHandling/SimpleEventBusTest.php
```diff
@@ -104,6 +104,41 @@ public function it_does_not_dispatch_new_events_before_all_listeners_have_run()
...
+    {
+        $domainMessage1 = $this->createDomainMessage(array('foo' => 'bar'));
+        $domainMessage2 = $this->createDomainMessage(array('foo' => 'bas'));
+
+        $domainEventStream1 = new DomainEventStream(array($domainMessage1));
+        $domainEventStream2 = new DomainEventStream(array($domainMessage2));
+
+        $eventListener = $this->createEventListenerMock();
+        $eventListener
+            ->expects($this->at(0))
```
cordoval: I hear from @sstok that a unit test that depends on internals is not good

asm89:
:+1: I tried to restart travis, not sure what's up there

cordoval:
it is just taking time https://travis-ci.org/qandidate-labs/broadway/jobs/50219658

test/Broadway/EventHandling/SimpleEventBusTest.php
kimlai: Would you mind explaining a bit ? I don't think I undersand what you mean (and I must admit I just re-used the other tests implementation here).

sstok: In this case it would be acceptable, but normally its better not make your test depend on the internal flow of a method your are testing.

The "problem" with PHPUnit mocks is that every method call made on a mock with `at()` will make the mock expect exactly that method call at that time. But if you change the calling order (later/sooner) then the test will fail the expectation. So every time you do a refactor the tests need to be updated and can't be safely used to test if you broke something.

As this interface has only one method its less fragile, and therefor less likely to break.
But if you have an interface with multiple methods, and you add expectations for all of them in a single test its going to break someday.

kimlai: Ah I see. I vaguely recall everzet saying something similar in a talk advocating for the use of Prophecy in place of PHPUnit mocks, but since I never experienced this pain first hand it didn't really stick.

I agree that the interface of listeners mitigate this greatly in this case. But thanks for the insight :).

wjzijderveld:
Thanks @kimlai!

PR #136, opened 2015-02-10
[Bundle] Make event store uuid column configurable.

kimlai:
With this change, one can configure the type of the uuid column in the
`DbalEventStore`.

The configuration option is
```
broadway:
    event_store:
        dbal:
            uuid_type: guid
```
Possible values are `guid` (default), `string` or `binary`.

It replaces the `useBinary` config option, but keeps the validation if the `binary` type is used.

I wrote this PR quickly just to know if that's something you would be interested in. My use case is that I use `string` uuids in my test suite fixtures, which makes debugging a bit easier, and tests more readable.

### DIFF
--- a/src/Broadway/Bundle/BroadwayBundle/DependencyInjection/BroadwayExtension.php
+++ b/src/Broadway/Bundle/BroadwayBundle/DependencyInjection/BroadwayExtension.php
@@ -104,8 +104,8 @@ class BroadwayExtension extends Extension
         );
 
         $container->setParameter(
-            'broadway.event_store.dbal.use_binary',
-            $config['dbal']['use_binary']
+            'broadway.event_store.dbal.uuid_type',
+            $config['dbal']['uuid_type']
         );
     }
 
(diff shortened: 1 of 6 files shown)


cordoval:
could we create a document upgrade or changelog to describe how to upgrade or take advantage of these changes?

kimlai:
Sure. But I'd like to know if that's something that is likely to be merged before I continue (upgrade documents, keeping BC, cleaning the code...).

wjzijderveld:
@kimlai I'm not sure about this feature... For most platforms (all except Postgres and SQLServer) Doctrine DBAL already uses a varchar under water, so I don't really see the added value to add this as a feature which is only used for testing.
But maybe I'm missing something crucial?

@asm89 What do you think?

kimlai:
> But maybe I'm missing something crucial?

Nope. I just have some end-to-end tests that need some Postgresql features, so I wanted to use pg for the event store in my test env, with my non-uuid uuids. But I hadn't realized that I could just configure separate sqlite dbal connection for the event store.

I guess you're gonna see that PR to stop hard-coding the dbal connection name soon :)

mbadolato:
> I guess you're gonna see that PR to stop hard-coding the dbal connection name soon :)

:+1:

kimlai:
Closing this, since https://github.com/qandidate-labs/broadway/pull/137 fixed my problem.

PR #146, opened 2015-03-05
Event sourced multiple dyanmic child entities example

simensen:
Hopefully the general idea of this example makes sense and is easy to follow. Someone looking for a job ( Job Seeker ) can claim to have held Jobs at some point. Jobs have a title and description and can be described again after they've been added. The idea being that the *job itself* was not changed, but only the description.

This shows how one can dynamically manage child entities for an AR including creating them, updating them (and the tricks involved there) and then removing them.

### DIFF
--- a/examples/event-sourced-multiple-dyanmic-child-entities/README.md
+++ b/examples/event-sourced-multiple-dyanmic-child-entities/README.md
@@ -36,7 +36,7 @@ moved correctly.
 Running Tests
 -------------
 
-The PHPUnit tests can be run by changing to the directory of the tests and running:
+The PHPUnit tests can be run by changing to this directory and running:
 
 ```bash
 $ phpunit .
--- a/examples/event-sourced-multiple-dyanmic-child-entities/README.md
+++ b/examples/event-sourced-multiple-dyanmic-child-entities/README.md
@@ -9,6 +9,33 @@ suite to test the domain.
 
 The two files contain comments about what is happening.
 
+
+The Story
+---------
+
+The purpose of this example is to show how dynamically created Child Entities of
+an Aggregate Root can be managed. Compare this to a Child Entity whose entire
+lifecycle is completely tied to its parent Aggregate Root where it is created
+when the Aggregate Root itself is created.
+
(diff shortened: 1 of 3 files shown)


jaymecd:
:+1:

wjzijderveld:
Could you add your description about the example in the README?
When I read the code for the first time it wasn't directly clear to me.

wjzijderveld:
I think code-wise this is fine. 

As we discussed on IRC a few times, I'm not convinced the examples are that clear when they get this big. But for now it's better then nothing :smile: 

We will probably have proper documentation someday......

simensen:
I'll add more description to the read me. :)

simensen:
@wjzijderveld I've updated this with more of a story. Is this what you were looking for?

examples/event-sourced-multiple-dyanmic-child-entities/README.md
```diff
@@ -0,0 +1,50 @@
...
+
+The tests show how the domain can be exercised using commands. It also shows
+how to test to ensure that Child Entities are being created, managed, and
+moved correctly.
+
+
+Running Tests
+-------------
+
+The PHPUnit tests can be run by changing to the directory of the tests and running:
```
wjzijderveld: `to the directory of the example`, running this from `/test` isn't going to work :smile:

wjzijderveld:
Minor comment, but otherwise :+1:

examples/event-sourced-multiple-dyanmic-child-entities/README.md
simensen: This is the same wording as all of the other tests. I can do a separate PR to fix them all if you like.

> The PHPUnit tests can be run by changing to this directory and running:

If you want me to change all of the README's here in this PR I can do that as well.

simensen:
@wjzijderveld ok, i made the requested change. i'm sending another pr for the other examples that use the same wording.

wjzijderveld:
Thanks!

simensen:
@wjzijderveld Thanks for your patience w/ me on this. :)

Issue #111, opened 2014-12-30
Snapshotting support


F21:
It would be nice if there was snapshotting support so that we can snapshot and cache a representation of a domain so that we don't have to build it by replaying events.

A storage interface would be nice as well, so that implementers can implement a snapshot storage adapter for desired storages.

kimlai:
Hello,

I think I've reached the point where I need snapshots. I'm currently managing aggregates with up to 6401 events, and it can take more then 10 seconds to load them. Those are very long lived aggregates, and it seems to me that snapshots are unavoidable.

Here's roughly how I see things : 

```php
class Snapshot implements Serializable
{
    private $lastEventPlayhead;
    private $payload;
    //getters and serialization/deserialization...
}

interface SnapshotStore
{
    public function append(Snapshotable $aggregate);
    /** @return Snapshot */
    public function loadLatestSnapshot($uuid);
}

interface Snapshotable
{
    public static createFromSnapshot(Snapshot $s);
    /** @return Snapshot */
    public function takeSnapshot();
}

class EventSourcingRepository
{
    public function load($uuid)
    {
        $latestSnapshot = $this->snapshotStore->loadLatestSnapshot($uuid);
        $latestEvents = $this->eventStore->loadStartingFrom($snapshot->getLatestEventPlayhead());
        
        return $this->aggregateFactory->create($this->aggregateClass, $domainEventStream, $latestSnapshot);
    }
}

class SnapshotService
{
    private $snapshotStore;
    
    public function takeSnapshot(Snapshotable $aggregate)
    {
        $s = $aggregate->takeSnapshot();
        $this->snapshotStore->append($s);
    }
}
```

jaymecd:
:+1: for snapshot support

othillo:
To understand snapshotting, I've created my own (opinionated) POC: https://github.com/othillo/broadway-snapshotting.

In order for anyone to create their own snapshotting implementation (or maybe even an "official" broadway/snapshotting implementation) we need to be able to query the event store for events that have been recorded after a given playhead.

Reading this discussion I think the best approach is to add a loadFromPlayhead to the EventStoreInterface. (BC break). I can create a separate PR for this.

Does anyone want to help out on a newly created broadway/snapshotting repo?

francescotrucchia:
@othillo in our company we need the snapshot and I opened an internal card last week. So if you want, I'm in :-)

othillo:
thank you, @francescotrucchia !

[...]

PR #147, opened 2015-03-17
Add Snapshotting support for event sourced aggregates

kimlai:
This Pull Request tries to address https://github.com/qandidate-labs/broadway/issues/111.

Snapshots are regular Domain Messages, but they are stored in a separate store, for the reasons described [here](https://groups.google.com/forum/#!msg/event-store/W4PnHcVFhDM/NCzjQkeJZXAJ).

The aggregate is responsible for generating snapshots on demand, and reconstitutig itself from a snapshot (ie implementing a `apply[MySnapshotClass]` method).
When a repository retrieves an agrgegate, it will first try to load the latest snapshot, then the events that were recorded after the snapshot was taken and pass those to the aggregate factory.

I've kept backward compatibilty.

### DIFF
--- a/src/Broadway/EventSourcing/AggregateFactory/AggregateFactoryInterface.php
+++ b/src/Broadway/EventSourcing/AggregateFactory/AggregateFactoryInterface.php
@@ -3,6 +3,7 @@
 namespace Broadway\EventSourcing\AggregateFactory;
 
 use Broadway\Domain\DomainEventStreamInterface;
+use Broadway\Domain\DomainMessage;
 
 /**
  * Interface AggregateFactoryInterface
@@ -15,5 +16,5 @@ interface AggregateFactoryInterface
      *
      * @return \Broadway\EventSourcing\EventSourcedAggregateRoot
      */
-    public function create($aggregateClass, DomainEventStreamInterface $domainEventStream);
+    public function create($aggregateClass, DomainEventStreamInterface $domainEventStream, DomainMessage $snapshot);
 }
--- a/src/Broadway/EventStore/DBALEventStore.php
+++ b/src/Broadway/EventStore/DBALEventStore.php
@@ -20,6 +20,7 @@ use Broadway\Serializer\SerializerInterface;
 use Doctrine\DBAL\Connection;
 use Doctrine\DBAL\DBALException;
 use Doctrine\DBAL\Schema\Schema;
+use Doctrine\DBAL\Statement;
 use Doctrine\DBAL\Version;
 use Rhumsaa\Uuid\Uuid;
 
@@ -38,6 +39,8 @@ class DBALEventStore implements EventStoreInterface
     private $metadataSerializer;
 
     private $loadStatement = null;
+
+    private $loadLastStatement = null;
 
     private $tableName;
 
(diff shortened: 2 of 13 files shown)


wjzijderveld:
Nice, thanks for taking the time for this PR, I'm still a bit swamped at the moment, but really want to discuss this in more detail soonish.

src/Broadway/EventStore/DBALEventStore.php
```diff
@@ -69,10 +72,41 @@ public function __construct(
...
         $statement = $this->prepareLoadStatement();
         $statement->bindValue('uuid', $this->convertIdentifierToStorageValue($id));
+        $statement->bindValue('playhead', $playhead);
+        $statement->execute();
+
+        $events = $this->executeStatement($statement);
+
+        if (empty($events)) {
+            throw new EventStreamNotFoundException(sprintf('EventStream not found for aggregate with id %s', $id));
+        }
```
rjkip: This exception is thrown when no events have been appended since the last snapshot.

kimlai: Oh jeez you're right. I guess it should only be thrown when you're trying to load *all* events of an aggregate (ie from playhead 0).

rjkip: It would be good to test that aggregates can be loaded with only a snapshot and no new events.

The correct behaviour for `EventStore#load()` is to throw when there are truly no events for that ID in the event stream. In case of `$playhead !== 0`, this means it should actually check if there are no events at all for that ID. No events is not an exceptional case per se. Strictly, this would mean an extra query when no events are found and `$playhead !== 0`. I am not suggesting you implement this, I am merely outlining what is correct behaviour imo. Let's let the Qandidate guys take a look first.

F21:
Any updates on this one?

wjzijderveld:
I think we need to aggregate discriminator in the event store before we can provide an acual snapshotting solution. 

Besides from that, there are already a few things that I would like to discuss:
- Should we have an interface for event stores that support snapshotting? Lowers the barrier for new event stores a bit and helps with type-hinting in the Repository f.e.
- Should the AggregateRoot expose if it supports snapshotting? There is now a (small) edge case that an aggregate might not have implemented the method to apply the Snapshot event while there is a snapshot somehow. I doubt if it will actually cause issues (especially with the aggregate discriminator)...

src/Broadway/EventStore/DBALEventStore.php
wjzijderveld: I don't really think the extra query is worth the trouble. How could there ever be no events when you load from a snapshot? Then you would have to created a snapshot for an aggregate without events (which in normal cases is already impossible), right?

src/Broadway/EventSourcing/AggregateFactory/AggregateFactoryInterface.php
```diff
@@ -15,5 +16,5 @@
      *
      * @return \Broadway\EventSourcing\EventSourcedAggregateRoot
      */
-    public function create($aggregateClass, DomainEventStreamInterface $domainEventStream);
+    public function create($aggregateClass, DomainEventStreamInterface $domainEventStream, DomainMessage $snapshot);
```
jaymecd: I suggest to use concrete `Snapshot` class for this purpose, as `DomainMessage` could lead to misuse.

bouke-nederstigt:
What's the status on this? I'd be happy to help, but not quite sure what needs to be done.

othillo:
I'm closing this, see discussion in https://github.com/qandidate-labs/broadway/issues/111.

PR #150, opened 2015-04-06
added setter for subscribers

cordoval:
this will be important in replay, so that we can specifically set who is going to listen to the events to be replayed :older_man:

### DIFF
--- a/src/Broadway/EventHandling/SimpleEventBus.php
+++ b/src/Broadway/EventHandling/SimpleEventBus.php
@@ -57,4 +57,9 @@ class SimpleEventBus implements EventBusInterface
             }
         }
     }
+
+    public function setSubscribers(array $eventListeners)
+    {
+        $this->eventListeners = $eventListeners;
+    }
 }


wjzijderveld:
I don't really see the added benefit of this at the moment.
You can already add listeners, why do we need another method to add them in batch?

cordoval:
The reasons is simple. Once an event bus is configured, all listeners are in place and that is fine. All has been done through configuration and say in a symfony context the container is all compiled and ready to go.

However it is the case in a dynamic environment where you want to disable certain listeners dynamically without having to configure a new symfony environment or hack a compiler pass to rebuild the container just for this change.

This setter then allows you to grab the service and in situ provide the listeners that should exclusively be plugged into the bus such as in a specific replay or partial reconstruction of the read model. The setter is a commodity and will not harm at all the encapsulation of the class.

wjzijderveld:
Not yet.

The point is, that I don't really see the problem. This solution feels a bit like a hack, where you want to reuse an event bus instance, because you don't want to configure/build a new event bus with only the listeners you need. Or am I missing something?

cordoval:
your assessment is correct. However, when you do a replay of events, you publish the stream say sequentially. You would do it with the default kernel, in which you have yeah perhaps the bus that you use commonly and the other bus that you just mentioned that is only for this listener. They will run in parallel unless you shut down the common one. You must shut down but then there is no switch in particular helpful for this, this is the reason you want to ensure either you wipe out the common bus of subscribers and add the listener to the latter bus, or you just wipe the listeners and add this particular listener on that single bus. In either case there is no simple way to publish the stream without affecting the other listeners because you are replaying.

cordoval:
closing as lack of response. I have overridden this already but was trying to contribute. Is cool, just that I don't like PRs lingering if they are rejected.

PR #151, opened 2015-04-08
MongoDB as EventStore

jaymecd:
Introducing support of MongoDB as EventStore.

Features included in this PR:
- read queries against primary nodes for strong data consistency
- read committed isolation level to avoid reading partial event commits
- test cases adopted for MongoDB

### DIFF
--- a/src/Broadway/EventStore/MongoDBEventStore.php
+++ b/src/Broadway/EventStore/MongoDBEventStore.php
@@ -1,0 +1,371 @@
+<?php
+
+/*
+ * This file is part of the broadway/broadway package.
+ *
+ * (c) Qandidate.com <opensource@qandidate.com>
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+namespace Broadway\EventStore;
+
+use Broadway\Domain\DateTime;
+use Broadway\Domain\DomainEventStream;
+use Broadway\Domain\DomainEventStreamInterface;
+use Broadway\Domain\DomainMessage;
+use Broadway\Domain\Metadata;
+use Broadway\EventStore\Exception\InvalidIdentifierException;
+use Broadway\Serializer\SerializerInterface;
+use Broadway\UuidGenerator\UuidGeneratorInterface;
+use Doctrine\MongoDB\Collection;
+use Doctrine\MongoDB\Database;
+use Exception;
+use MongoClient;
+use MongoCollection;
+use MongoException;
+
+/**
+ * Event store using a relational database as storage.
+ *
+ * The implementation uses doctrine MongoDB for the communication with the
+ * underlying data store.
+ */
+class MongoDBEventStore implements EventStoreInterface
+{
+    const FIELD_STREAM_ID   = 'sid';
+    const FIELD_COMMIT_ID   = 'cid';
+    const FIELD_PLAYHEAD    = 'head';
+    const FIELD_PAYLOAD     = 'data';
+    const FIELD_METADATA    = 'meta';
+    const FIELD_TYPE        = 'type';
+    const FIELD_RECORDED_ON = 'time';
+
+    /**
+     * @var Collection
+     */
+    private $eventCollection;
+
+    /**
+     * @var Collection
+     */
+    private $transactionCollection;
+
+    /**
+     * @var SerializerInterface
+     */
+    private $payloadSerializer;
+
+    /**
+     * @var SerializerInterface
+     */
+    private $metadataSerializer;
+
+    /**
+     * @var UuidGeneratorInterface
+     */
+    private $uuidGenerator;
+
+    /**
+     * @param Database               $database
+     * @param SerializerInterface    $payloadSerializer
+     * @param SerializerInterface    $metadataSerializer
+     * @param UuidGeneratorInterface $uuidGenerator
+     * @param string                 $eventCollectionName
+     * @param string                 $transactCollectionName
+     */
+    public function __construct(
+        Database $database,
+        SerializerInterface $payloadSerializer,
+        SerializerInterface $metadataSerializer,
+        UuidGeneratorInterface $uuidGenerator,
+        $eventCollectionName,
+        $transactCollectionName
+    ) {
+        $this->payloadSerializer    = $payloadSerializer;
+        $this->metadataSerializer   = $metadataSerializer;
+        $this->uuidGenerator        = $uuidGenerator;
+
+        $this->eventCollection = $database->selectCollection($eventCollectionName);
+        $this->eventCollection->setReadPreference(MongoClient::RP_PRIMARY);
+
+        $this->transactionCollection = $database->selectCollection($transactCollectionName);
+        $this->transactionCollection->setReadPreference(MongoClient::RP_PRIMARY);
+    }
+
+    /**
+     * {@inheritDoc}
+     */
+    public function load($id)
+    {
+        $cursor = $this->eventCollection
+            ->find(array(
+                self::FIELD_STREAM_ID => (string) $id,
+                self::FIELD_COMMIT_ID => null,
+            ), array(
+                '_id' => false,
+            ))
+            ->sort(array(self::FIELD_PLAYHEAD => MongoCollection::ASCENDING))
+        ;
+
+        $events = array();
+
+        foreach ($cursor as $row) {
+            $events[] = $this->deserializeEvent($row);
+        }
+
+        $cursor = null;
+
+        if (!$events) {
+            throw new EventStreamNotFoundException(sprintf('EventStream not found for aggregate with id %s', $id));
+        }
+
+        return new DomainEventStream($events);
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    public function append($id, DomainEventStreamInterface $eventStream)
+    {
+        $aggregateId = (string) $id;
+
+        $events = $this->extractEvents($eventStream, $aggregateId);
+
+        if (!$events) {
+            return;
+        }
+
+        try {
+            if (1 == count($events)) {
+                $this->appendEvent(reset($events));
+            } else {
+                $this->appendEvents($events);
+            }
+        } catch (MongoException $exception) {
+            throw MongoDBEventStoreException::create($exception);
+        } catch (Exception $exception) {
+            throw new MongoDBEventStoreException($exception->getMessage(), 0, $exception);
+        }
+    }
+
+    /**
+     * Configure collection indices. Should be used by migration or by bootstrapping tests.
+     */
+    public function configureCollection()
+    {
+        $this->eventCollection->ensureIndex(array(
+            self::FIELD_STREAM_ID => MongoCollection::ASCENDING,
+            self::FIELD_PLAYHEAD  => MongoCollection::ASCENDING,
+        ), array(
+            'unique'     => true,
+            'background' => true,
+        ));
+    }
+
+    /**
+     * @param DomainEventStreamInterface|DomainMessage[] $eventStream
+     * @param string                                     $aggregateId
+     *
+     * @return array
+     */
+    private function extractEvents(DomainEventStreamInterface $eventStream, $aggregateId)
+    {
+        $events = array();
+
+        foreach ($eventStream as $message) {
+            $this->assertIdentifier($aggregateId, $message->getId());
+
+            $events[] = $this->serializeEvent($message);
+        }
+
+        return $events;
+    }
+
+    /**
+     * @param string $expected
+     * @param string $actual
+     */
+    private function assertIdentifier($expected, $actual)
+    {
+        if (0 != strcmp($expected, $actual)) {
+            throw new InvalidIdentifierException(sprintf('Expecting %s, got %s identifier.', $expected, $actual));
+        }
+    }
+
+    /**
+     * @param DomainMessage $message
+     *
+     * @return array
+     */
+    private function serializeEvent(DomainMessage $message)
+    {
+        $data = array(
+            self::FIELD_STREAM_ID   => (string) $message->getId(),
+            self::FIELD_PLAYHEAD    => $message->getPlayhead(),
+            self::FIELD_METADATA    => $this->metadataSerializer->serialize($message->getMetadata()),
+            self::FIELD_PAYLOAD     => $this->payloadSerializer->serialize($message->getPayload()),
+            self::FIELD_RECORDED_ON => $message->getRecordedOn()->toString(),
+            self::FIELD_TYPE        => $message->getType(),
+        );
+
+        // save space
+        if (empty($data[self::FIELD_METADATA])) {
+            unset($data[self::FIELD_METADATA]);
+        }
+
+        return $data;
+    }
+
+    /**
+     * @param array $row
+     *
+     * @return DomainMessage
+     */
+    private function deserializeEvent(array $row)
+    {
+        if (empty($row[self::FIELD_METADATA])) {
+            $metadata = new Metadata();
+        } else {
+            $metadata = $this->metadataSerializer->deserialize($row[self::FIELD_METADATA]);
+        }
+
+        return new DomainMessage(
+            $row[self::FIELD_STREAM_ID],
+            $row[self::FIELD_PLAYHEAD],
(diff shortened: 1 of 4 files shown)


src/Broadway/EventStore/MongoDBEventStore.php
```diff
@@ -0,0 +1,367 @@
...
+
+        $events = $this->extractEvents($eventStream, $aggregateId);
+
+        if (!$events) {
+            return;
+        }
+
+        if (count($events) == 1) {
+            try {
+                $this->saveEvent(reset($events));
```
wjzijderveld: Is there a specific reason you choose to do this separately if there is exactly 1 event? I'm not really seeing the benefit.

src/Broadway/EventStore/MongoDBEventStore.php
```diff
@@ -0,0 +1,367 @@
...
+            throw MongoDBEventStoreException::create($exception);
+        } catch (Exception $exception) {
+            throw new MongoDBEventStoreException($exception->getMessage(), 0, $exception);
+        }
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    public function configureCollection()
```
wjzijderveld: As it is only used in the test at the moment, do we really need this?

src/Broadway/EventStore/MongoDBEventStore.php
jaymecd: Single document insert does not require 2pc. It's done to save number of requests.

src/Broadway/EventStore/MongoDBEventStore.php
jaymecd: It's done in the same way as DBAL implementation, it requires indices to be set before processing.

src/Broadway/EventStore/MongoDBEventStore.php
```diff
@@ -0,0 +1,367 @@
...
+     * {@inheritDoc}
+     */
+    public function load($id)
+    {
+        $cursor = $this->eventCollection
+            ->find(array(
+                self::FIELD_STREAM_ID => (string) $id,
+                self::FIELD_COMMIT_ID => null,
+            ), array(
+                '_id' => 0,
```
wjzijderveld: Please use `true`/`false` instead of 0/1

src/Broadway/EventStore/MongoDBEventStore.php
```diff
@@ -0,0 +1,367 @@
...
+    }
+
+    /**
+     * @param array $row
+     *
+     * @return DomainMessage
+     */
+    private function deserializeEvent(array $row)
+    {
+        if (empty($row[self::FIELD_METADATA])) {
```
wjzijderveld: As far as I can see, it's not possible to save empty (or no) metadata, so just always deserialize it

src/Broadway/EventStore/MongoDBEventStore.php
```diff
@@ -0,0 +1,367 @@
...
+        );
+    }
+
+    /**
+     * @param array  $event
+     * @param string $commitId
+     */
+    private function saveEvent(array $event, $commitId = null)
+    {
+        if ($commitId) {
```
wjzijderveld: Please remove this here.
Just make sure the field is set when you need it.

src/Broadway/EventStore/MongoDBEventStore.php
```diff
@@ -0,0 +1,367 @@
...
+     * @throws MongoException
+     */
+    private function saveBulkEvents(array $events)
+    {
+        $commitId = $this->getNewId();
+
+        try {
+            $this->beginTransaction($commitId);
+
+            foreach ($events as $event) {
```
wjzijderveld: Why not use [batchInsert](http://php.net/manual/en/mongocollection.batchinsert.php) here?

src/Broadway/EventStore/MongoDBEventStore.php
```diff
@@ -0,0 +1,367 @@
...
+        $exception = null;
+
+        try {
+            $this->eventCollection->remove(array(
+                self::FIELD_COMMIT_ID => $commitId,
+            ), array(
+                'safe' => true,
+            ));
+        } catch (MongoException $e) {
+            $exception = $exception ?: $e;
```
wjzijderveld: `$exception` won't ever be set here, please just use `$exception = $e`;

src/Broadway/EventStore/MongoDBEventStore.php
```diff
@@ -0,0 +1,367 @@
...
+
+    /**
+     * @var Collection
+     */
+    private $eventCollection;
+
+    /**
+     * @var Collection
+     */
+    private $transactCollection;
```
wjzijderveld: More verbose is more better :)
Please make this `$transactionCollection`

src/Broadway/EventStore/MongoDBEventStore.php
```diff
@@ -0,0 +1,367 @@
...
+            $this->transactCollection->remove(array(
+                '_id' => $commitId,
+            ), array(
+                'safe' => true,
+            ));
+        } catch (MongoException $e) {
+            $exception = $exception ?: $e;
+        }
+
+        if ($exception) {
```
wjzijderveld: `if (null !== $exception)`

src/Broadway/EventStore/MongoDBEventStore.php
jaymecd: It depends on serialization. Eg, save only payload, as class name is known and never to be changed. Or why save empty payload of metadata, imagine overhead of 100M docs.

wjzijderveld:
Thanks for the PR, really appreciate it!

src/Broadway/EventStore/MongoDBEventStore.php
jaymecd: Good catch, missed that.

jaymecd:
@wjzijderveld updated, but scrutinizer failed by some unexpected reason. 
Any idea? I've contacted support, waiting.

wjzijderveld:
I just restarted The scrutrnizer job, they fsil from time to time.

I'll look at you other comments later :-)

src/Broadway/EventStore/MongoDBEventStore.php
```diff
@@ -0,0 +1,371 @@
...
+            }
+        } catch (MongoException $exception) {
+            throw MongoDBEventStoreException::create($exception);
+        } catch (Exception $exception) {
+            throw new MongoDBEventStoreException($exception->getMessage(), 0, $exception);
+        }
+    }
+
+    /**
+     * {@inheritdoc}
```
wjzijderveld: This method isn't in the interface, no doc-block required here.

jaymecd: updated

francescotrucchia:
:+1:

wjzijderveld:
@qandidate-labs/broadway I'm fine with merging this as is (can optionally fix the exception assignment separate). What do others think?

@jaymecd Have you been running this in production? If not, maybe we should add a note in the README that this is experimental?

jaymecd:
@wjzijderveld nope, not yet, so `experimental` would be important.

src/Broadway/EventStore/MongoDBEventStore.php
```diff
@@ -0,0 +1,371 @@
...
+            throw new MongoDBEventStoreException($exception->getMessage(), 0, $exception);
+        }
+    }
+
+    /**
+     * Configure collection indices. Should be used by migration or by bootstrapping tests.
+     */
+    public function configureCollection()
+    {
+        $this->eventCollection->ensureIndex(array(
```
prolic: ensureIndex is deprecated. use createIndex instead.

prolic: sorry, doctrine mongodb uses still the ensureIndex method ;)

src/Broadway/EventStore/MongoDBEventStore.php
```diff
@@ -0,0 +1,371 @@
...
+        );
+    }
+
+    /**
+     * @param array $event
+     */
+    private function appendEvent(array $event)
+    {
+        $this->eventCollection->insert($event, array(
+            'safe' => true,
```
jmikola: Acknowledged writes have been the default since driver version 1.3, so this shouldn't be necessary. If you did want to specify the option, `['w' => 1]` would be the equivalent option to use (`safe` is deprecated).

I'm not sure how important the data being logged by this service is, but you may want `['w' => 'majority']` when working with a replica set (in that case, it'd probably be best to make the write concern a configurable option).

src/Broadway/EventStore/MongoDBEventStore.php
```diff
@@ -0,0 +1,371 @@
...
+     */
+    private function commitEvents($commitId, array $events)
+    {
+        foreach ($events as &$event) {
+            $event[self::FIELD_COMMIT_ID] = $commitId;
+        }
+        
+        unset($event);
+
+        $this->eventCollection->batchInsert($events, array(
```
jmikola: `batchInsert` is also deprecated in favor of the [MongoWriteBatch](http://php.net/mongowritebatch) classes, which were added in driver version 1.5 and supported by MongoDB 2.6+. Depending on whether those version requirements are acceptable, I'd suggest swapping this over.

src/Broadway/EventStore/MongoDBEventStore.php
```diff
@@ -0,0 +1,371 @@
...
+use Broadway\Domain\DateTime;
+use Broadway\Domain\DomainEventStream;
+use Broadway\Domain\DomainEventStreamInterface;
+use Broadway\Domain\DomainMessage;
+use Broadway\Domain\Metadata;
+use Broadway\EventStore\Exception\InvalidIdentifierException;
+use Broadway\Serializer\SerializerInterface;
+use Broadway\UuidGenerator\UuidGeneratorInterface;
+use Doctrine\MongoDB\Collection;
+use Doctrine\MongoDB\Database;
```
jmikola: Is there any value in using the Doctrine classes instead of the raw driver? This class seems comparable to the session and cache adapters I worked on for Symfony and Zend, since it consists of just a few direct read and write ops.

As noted below, Doctrine doesn't yet support the MongoWriteBatch classes (for batch ops via the write commands in 2.6+), nor the `createIndex()` method (although the deprecated `ensureIndex()` method is an alias). Were you to make the switch, I'd suggest requiring at least `>=1.5.0` of the `mongo` extension (although it wouldn't hurt to bump to `1.6.0` given that this is a new integration).

jaymecd:
@jmikola thn for tips, I review code a bit later.

jmikola:
@jaymecd: No problem. @asm89 just sent me this PR over IRC and invited me to take a look :)

danizord:
Any news on this one?

prolic:
MongoDB has no transaction support. Using the [$isolated](http://docs.mongodb.org/manual/reference/operator/update/isolated/) operator you can achieve atomic updates on a single collection, but not on multiple ones. Additionally the $isolated operator does not work in a sharded cluster environment. That said, it's not a good idea to run this in a very high performant situation with sharding enabled.
You can check out our implementation of the [mongo-db-adapter](https://github.com/prooph/event-store-mongodb-adapter) for [prooph/event-store](https://github.com/prooph/event-store), to see how this is working with manual transaction support. The PR to address the $isolated operator and changing README notes is here: https://github.com/prooph/event-store-mongodb-adapter/pull/27.

jaymecd:
Hey guys, I spent last weeks in hospital.
Continue with PR in next days.
Tnx for understanding.

wjzijderveld:
@jaymecd No problem, hope you are okay!

If you want any help on it, let me know!

jaymecd:
@wjzijderveld, I've made a deep dive into the problems and issues noted by @jmikola and @prolic and have doubts if solution should be based on: 

- [mongo extension](http://pecl.php.net/package/mongo) (legacy)
  - recently marked as deprecated
  - used by many
- [mongodb extension](http://pecl.php.net/package/mongodb) 
  - supersedes `mongo` extension
  - supports both `php` and `hhvm`
  - has own abstraction library [mongodb/mongo-php-library](https://github.com/mongodb/mongo-php-library)
  - `NB!` do not support [legacy-based integration](https://github.com/mongodb/mongo-php-driver/blob/master/README.md#peclmongodb-mongodb-driver-for-php), eg. `Doctrine MongoDB`, due to the **different API**

So, I'm asking for your opinion on this issue before continue. Personally I'd go with `mongodb` extension/library and then, if it's required, with legacy extension support as separate storage class `MongoDBLegacyEventStore`.

wjzijderveld:
@jaymecd Thanks for diving into it!

Hmm, tricky choice there. But I think the better way would be to go with the new extension. Compatibility with HHVM is something I would definitely like to support :-)
Downsides: we use the deprecated mongo extension (with Doctrine) for Saga states at the moment. So we probable should do something about that as well.

@qandidate-labs/broadway What do you think?

codeliner:
Also the legacy driver will not be updated to support php7. No HHVM support ok, but PHP7 should not be a question ;) If you also want to have legacy version you can wrap the 1.0 version of [prooph/event-store-mongodb-adapter](https://github.com/prooph/event-store-mongodb-adapter/releases/tag/v1.0) in a class implementing Broadway event store. Maybe faster than writing a legacy adapter yourself.
Our 2.0 version will use the new extension.

jaymecd:
@codeliner, I don't think wrapping `prooph` implementation would be a good idea, as it brings another level to `diamond dependencies` + `prooph/common`: one ES implementation partially includes another one - that's more than strange.

jaymecd:
@codeliner @jmikola - [$isolated](https://docs.mongodb.org/manual/reference/operator/update/isolated/) operator does not work in a sharded cluster.

Is there any way to have same atomicity within sharded setup? As event store data could consume all disk space on single shard, isn't it?

codeliner:
@jaymecd : @prolic is our mongoDB specialist. As far as I know there is no way. We are looking for alternatives, too. Depending on your domain contexts one shard per context is maybe enough. Also when using snapshots in combination with "archiving" old events you may not run out of disk space. So mongoDB is still a valid option in my eyes.

> one ES implementation partially includes another one

Well, Broadway provides only an ES interface and different independent ES implementations. prooph/event-store on the other hand uses different adapters. I did not say you should require prooph/event-store but only one of the available adapters. Our packages are stand-alone (btw. something that Broadway should consider, too). However, it was more a hint for people who may need to use the old extension but want to use other features of Broadway. I don't consider Broadway a competitor. Would like to use the Saga implementation but unfortunately it is not available as a stand-alone package and installing everything incl. the Sf bundle, aggregate root classes, etc. is really not an option.

prolic:
Running out of disk-space? Really? I can store all your events on my micro-sd card! Unless you are not twitter and have multiple thousands events per second, you never run out of disk space. Even if so, you should upgrade your hardware and buy a larger hard disk.

For $isolated: There is no other solution, you can't run mongodb event store in a sharded cluter environment. If you need more speed, consider using postgres.

sstok:
`sharded cluter` epic typo :joy: 

> If you need more speed, consider using postgres.
Here :+1:

jaymecd:
In order to save time I've started to update implementation (using hints above) with `mongodb` extension and it's library to have MongoDB as event store, as PHP7 & HHVM really worth it.

Konafets:
Hi @jaymecd,

> In order to save time I've started to update implementation (using hints above) with mongodb extension and it's library to have MongoDB as event store, as PHP7 & HHVM really worth it.

How is your progress on that?

wjzijderveld:
I started to refactor the Saga\MongoDBRepository to use the new extension #236

jaymecd:
@Konafets @wjzijderveld I didn't had a chance to continue on that, so far could not provide any ETA from my side. 
I would highly appreciate if you could continue on his PR.

othillo:
I'm closing this for now. Additional event stores are very welcome, though! We are actually splitting implementations to separate repositories at the moment. If someone wants to work on a new broadway/event-store-mongodb implementation, please ping us! /cc @broadway/broadway 

Thank you for your effort!

othillo:
work on thiswill continue in this repo: https://github.com/broadway/event-store-mongodb

PR #153, opened 2015-04-10
Fix string conversion of Saga name in thrown error message

mbadolato:
StaticallyConfiguredSagaMetadataFactory was not properly throwing a ```\RuntimeException()``` if a saga did not implement. Instead PHP was issuing a fatal error when creating the exception message to throw:

```
Object of class Path\To\MySaga could not be converted to string

/my/path/vendor/broadway/broadway/src/Broadway/Saga/Metadata/StaticallyConfiguredSagaMetadataFactory.php:27
```


wjzijderveld:
Tnx!

PR #155, opened 2015-04-17
Allow setting of authentication related settings for the elasticsearch client

richardmiller:
Currently only the hosts can be set but the elasticsearch client does not work with basic auth username and password passed in the url when using https. Whilst this is probably a bug with the elasticsearch client this change allows the relevant auth options to be passed to the elasticsearch client when using the Symfony bundle.


wjzijderveld:
Thanks!

PR #156, opened 2015-04-17
Check for error responses from elasticsearch

richardmiller:
I ran into an issue where we were getting 401 responses back from elasticsearch because of the issue in #155. These were getting swallowed though so that it looked as though everything was ok but that no results were found. 

Not sure is this is an ideal way of dealing with this though.

### DIFF
--- a/src/Broadway/ReadModel/ElasticSearch/ElasticSearchRepository.php
+++ b/src/Broadway/ReadModel/ElasticSearch/ElasticSearchRepository.php
@@ -62,7 +62,8 @@ class ElasticSearchRepository implements RepositoryInterface
             'refresh' => true,
         );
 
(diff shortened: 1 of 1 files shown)


wjzijderveld:
Not sure about this way as well. But besides that: I'm not sure what happened actually, because as far as I can see (haven't tried to reproduce yet), elasticsearch client already throws exceptions and we only catch the 404's.

richardmiller:
Will take another look, the elasticsearch client is not throwing an exception, so may well be a bug with that.

asm89:
@richardmiller Did you manage to find anything?

richardmiller:
Will close as I think this is an issue in the elasticsearch client

PR #157, opened 2015-04-20
Added comment about basic implementation

wjzijderveld:
Added some explanation about the read model repositories.

Addresses [some unclarities](https://github.com/qandidate-labs/broadway/pull/110) about the current implementation


asm89:
:shipit: :+1:

PR #163, opened 2015-05-16
replace rhumsaa/uuid by ramsey/uuid

BenoitLeveque:
it's the same repository behind.

this commit will prevent the line bellow to be display each times you install this lib

```
Package rhumsaa/uuid is abandoned, you should avoid using it. Use ramsey/uuid instead
```


wjzijderveld:
Thanks! Hadn't notices yet

cordoval:
:+1:

PR #164, opened 2015-05-18
[RFC] upcasting

ged15:
Upcasting is used when structure of data in events changes. This is an initial implementation. Heavily inspired by implementation in [Axon Framework](http://www.axonframework.org/docs/2.0/repositories-and-event-stores.html).

Todo:
- [ ] Documentation
- [ ] Bundle config

### DIFF
--- a/test/Broadway/EventSourcing/AbstractEventSourcingRepositoryTest.php
+++ b/test/Broadway/EventSourcing/AbstractEventSourcingRepositoryTest.php
@@ -25,11 +25,17 @@ use Broadway\EventStore\EventStoreInterface;
 use Broadway\EventStore\InMemoryEventStore;
 use Broadway\EventStore\TraceableEventStore;
 use Broadway\ReadModel\Projector;
+use Broadway\Serializer\SerializableInterface;
+use Broadway\Serializer\SimpleInterfaceSerializer;
 use Broadway\TestCase;
+use Broadway\Upcasting\SequentialUpcasterChain;
 use RuntimeException;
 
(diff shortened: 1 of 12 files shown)


test/Broadway/EventSourcing/AbstractEventSourcingRepositoryTest.php
```diff
@@ -25,11 +25,17 @@
...
 use Broadway\ReadModel\Projector;
+use Broadway\Serializer\SerializableInterface;
+use Broadway\Serializer\SimpleInterfaceSerializer;
 use Broadway\TestCase;
+use Broadway\Upcasting\SequentialUpcasterChain;
 use RuntimeException;
 
 abstract class AbstractEventSourcingRepositoryTest extends TestCase
 {
+    /** @var \PHPUnit_Framework_MockObject_MockObject */
```
sstok: Doesn't seem to be used.

TomasVotruba: Rather property, but yes.

wjzijderveld:
@ged15 Is it possible to add an example how to use this? (Which is probably the documentation part :smile: )

Besides that, I hoped we would use this in (pre-) production before merging this, but as it doesn't seem we will be needing this on the short-term, I would like to merge this sooner. But would like to see some examples how you would use this.

othillo:
I'm closing this because it has been open for too long.

An updated PR with documentation and examples is more than welcome!

PR #167, opened 2015-06-10
fixed deprecation warnings

ged15:
ok this doesn't work on Symfony 2.3.* :/

### DIFF
--- a/src/Broadway/Bundle/BroadwayBundle/Resources/config/saga/mongodb.xml
+++ b/src/Broadway/Bundle/BroadwayBundle/Resources/config/saga/mongodb.xml
@@ -9,21 +9,15 @@
             <argument type="service" id="broadway.saga.state.mongodb_collection" />
         </service>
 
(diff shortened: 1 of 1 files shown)


wjzijderveld:
Yeah, this definitely would be the easy fix. But as 2.3 still has support till May 2016 (and security patches till May 2017) we cannot just drop support for 2.3...

ged15:
Definitely. Already looking into ways to solve this. Also into ways to execute builds with Symfony 2.3 and the latest version.

wjzijderveld:
Cool, haven't got around doing that.

Most commonly used is something like this: https://github.com/hwi/HWIOAuthBundle/blob/master/.travis.yml#L15-L24

wjzijderveld:
As this is not a permanent solution and Symfony 'fixed' the warnings now, I'll close this PR.

I think we still need to come with a solution to support 3.0, but let's think about a proper solution there.

PR #177, opened 2015-06-26
AbstractEventSourcingRepositoryTest didnt use the correct class

wjzijderveld:
Ran into this when running tests separately, The TestEvenSourcedAggregate it is referring to is in `test/Broadway/EventSourcing/EventSourcingRepositoryTest.php`.


othillo:
:+1:

PR #178, opened 2015-06-27
Travis Build Matrix

wjzijderveld:
First try at a build matrix to test against Symfony 2.3 and 2.7
Can open a separate PR with the composer.lock file when needed. Tests fail way to often now without the lockfile.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,2 +1,1 @@
 /vendor/
-composer.lock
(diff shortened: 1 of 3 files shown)


wjzijderveld:
@qandidate-labs/broadway Anybody want to CR/merge this?

othillo:
:+1:

Issue #120, opened 2015-01-16
Loading by id of a different aggregate


richardmiller:
Sorry if this is a silly question and I am missing something. 

Since the same table is used for all events in the event store: if I have two aggregates classes and I call `EventSourcingRepository::load()` on a repository for one aggregate class with an id belonging to an instance of the other aggregate class then there is no exception. Instead the events are played back against the wring type of entity and silently do nothing as there are no methods to handle them. 

Am I (a) doing something wrong to make this possible at all (b) missing something that should make this throw an exception (c) just needing to implement something that stops this within my aggregates?

rjkip:
Hi @richardmiller. Yes, aggregate roots of different types but with the same ID appear to receive each other's events. It is assumed you're using UUIDs to generate IDs for your aggregate roots (as auto-increment is not feasible), and collisions are not likely to occur. I have used the DBAL event store and I enforce uniqueness across roots by creating a primary key of (uuid, playhead) on the event stream table.

richardmiller:
Hi, I'm using UUIDs - it's not a collision I'm worried about it. If I have a Car aggregate and a Garage aggregate and I have a uuid for a Car and accidentally use that uuid to load something from the Garage EventSourcingRepository then it will find the Car related events and try and replay them in a Garage aggregate instance.

rjkip:
Ah! Yes, you'd want an `AggregateNotFoundException`. This seems like an issue that lies in the `EventStoreInterface` rather than your aggregate (it should fail before that). Receiving a clean aggregate of a different type is unacceptable to me. @asm @wjzijderveld, any reason the event stream is not limited by the aggregate's class?

asm:
Hrm, I think you guys have the wrong @asm...

rjkip:
Heh, @asm89 :sweat_smile:

mbadolato:
> If I have a Car aggregate and a Garage aggregate and I have a uuid for a Car and accidentally use that uuid to load something from the Garage EventSourcingRepository then it will find the Car related events and try and replay them in a Garage aggregate instance.

I guess my question is, how is this really a problem? You're calling ```load($id)``` on a ``EventSourcingRepository``` object, which knows what type of aggregate class it is loading and reconstituting.  

[...]

PR #181, opened 2015-07-02
Stream type

wjzijderveld:
This is a start to add the stream type (in most cases the Aggregate Root) to the EventStore.

Still TODO:
- [ ] Check if we need to normalize the classname
- [x] Update CHANGELOG
- [ ] Give some hints/snippets to add a migration to your application
- [x] Provide a more sensible name, I don't really like stream type here

### DIFF
--- a/src/Broadway/EventStore/EventStoreInterface.php
+++ b/src/Broadway/EventStore/EventStoreInterface.php
@@ -26,9 +26,8 @@ interface EventStoreInterface
     public function load($streamType, $identifier);
 
     /**
-     * @param mixed                      $streamType
      * @param mixed                      $identifier
      * @param DomainEventStreamInterface $eventStream
      */
-    public function append($streamType, $identifier, DomainEventStreamInterface $eventStream);
+    public function append($identifier, DomainEventStreamInterface $eventStream);
 }
--- a/src/Broadway/EventStore/EventStoreInterface.php
+++ b/src/Broadway/EventStore/EventStoreInterface.php
@@ -26,8 +26,9 @@ interface EventStoreInterface
     public function load($streamType, $identifier);
 
     /**
+     * @param mixed                      $streamType
      * @param mixed                      $identifier
      * @param DomainEventStreamInterface $eventStream
      */
-    public function append($identifier, DomainEventStreamInterface $eventStream);
+    public function append($streamType, $identifier, DomainEventStreamInterface $eventStream);
 }
--- a/src/Broadway/EventStore/EventStoreInterface.php
+++ b/src/Broadway/EventStore/EventStoreInterface.php
@@ -23,11 +23,12 @@ interface EventStoreInterface
      *
      * @return DomainEventStreamInterface
      */
-    public function load($id);
+    public function load($streamType, $identifier);
 
     /**
-     * @param mixed                      $id
+     * @param mixed                      $streamType
+     * @param mixed                      $identifier
      * @param DomainEventStreamInterface $eventStream
      */
-    public function append($id, DomainEventStreamInterface $eventStream);
+    public function append($streamType, $identifier, DomainEventStreamInterface $eventStream);
 }
--- a/src/Broadway/EventStore/InMemoryEventStore.php
+++ b/src/Broadway/EventStore/InMemoryEventStore.php
@@ -44,21 +44,19 @@ class InMemoryEventStore implements EventStoreInterface, EventStoreManagementInt
     /**
      * {@inheritDoc}
      */
-    public function append($identifier, DomainEventStreamInterface $eventStream)
+    public function append($streamType, $identifier, DomainEventStreamInterface $eventStream)
     {
         $identifier = (string) $identifier;
 
-        foreach ($eventStream as $domainMessage) {
-            $streamType = $domainMessage->getStreamType();
-            $playhead   = $domainMessage->getPlayhead();
+        if (! isset($this->events[$streamType][$identifier])) {
+            $this->events[$streamType][$identifier] = array();
+        }
 
-            if (! isset($this->events[$streamType][$identifier])) {
-                $this->events[$streamType][$identifier] = array();
-            }
-
+        foreach ($eventStream as $event) {
+            $playhead = $event->getPlayhead();
             $this->assertPlayhead($this->events[$streamType][$identifier], $playhead);
 
-            $this->events[$streamType][$identifier][$playhead] = $domainMessage;
+            $this->events[$streamType][$identifier][$playhead] = $event;
         }
     }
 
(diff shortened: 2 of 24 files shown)


wjzijderveld:
This issues relates to #111 and #120

simensen:
Is there anything I can do to help this along?

wjzijderveld:
Not sure how you can help right now, I would like to have some more people involved to discuss this in more detail.. I'm just not sure this is __the__ solution, so I'm curious to other opinions and ideas.

jacobkiers:
On the plus side, this implementation brings Broadway closer to the implementation of a few other ES implementations, as described in #120. That in itself would make it useful, I think.

On the other hand, this breaks BC. I know we're not yet at 1.0, so technically that would not be a problem. In my experience however, breaking BC is _always_ painful, especially more so when it impacts data stores :grimacing:. So if we do go this route, I would like a more BC-friendly solution.

Personally, for us at Alphacomm, we just use the DBAL Event store with different tables for each aggregate. That solves the problem nicely, and the problem as described in #120 immediately goes away as well...

All in all, I'm moderately positive about the idea.

With regards to naming: why not just name it something like _`$aggregateType`_? Since AFAICS the Event Store is exclusively used to store events _for aggregates_, making that explicit is a good thing, I think.

wjzijderveld:
I'm not sure about aggregateType, personally I'm fine with it, but as aggregate is a term related to DDD and not event sourcing per se, I'm not sure it's the best fit.

@jacobkiers And I'm sorry, but BC breaks will happen... We deliberatly didn't tag a 1.0 yet for that specific reason, as we are aware that there will be more BC breaks coming.

We will tag 1.0 eventually, but I would like to have more people using it to find some issues that might still exist, especially in the newer stuff like Snapshotting and the event store management stuff.

src/Broadway/EventStore/EventStoreInterface.php
```diff
@@ -23,11 +23,12 @@
      *
      * @return DomainEventStreamInterface
      */
-    public function load($id);
+    public function load($streamType, $identifier);
```
asm89: `$streamType` :+1:  or just `$type` (you already know you're getting a `DomainEventStreamInterface`

src/Broadway/EventStore/InMemoryEventStore.php
```diff
@@ -26,33 +26,33 @@ class InMemoryEventStore implements EventStoreInterface
...
+        $identifier = (string) $identifier;
 
-        if (isset($this->events[$id])) {
-            return new DomainEventStream($this->events[$id]);
+        if (isset($this->events[$identifier])) {
+            return new DomainEventStream($this->events[$identifier]);
         }
 
-        throw new EventStreamNotFoundException(sprintf('EventStream not found for aggregate with id %s', $id));
+        throw new EventStreamNotFoundException(sprintf('EventStream not found for aggregate with id %s', $identifier));
```
asm89: "and type %s"

src/Broadway/EventStore/InMemoryEventStore.php
```diff
@@ -26,33 +26,33 @@ class InMemoryEventStore implements EventStoreInterface
     /**
      * {@inheritDoc}
      */
-    public function load($id)
+    public function load($streamType, $identifier)
```
asm89: This whole class should start storing events per `$streamType` or the like.

test/Broadway/EventSourcing/AbstractEventSourcingRepositoryTest.php
```diff
@@ -30,6 +30,8 @@
 
 abstract class AbstractEventSourcingRepositoryTest extends TestCase
 {
+    const AGGREGATE_CLASS = 'MYAggregate';
```
asm89: nit: `s/MY/My`

test/Broadway/EventStore/EventStoreTest.php
```diff
@@ -21,6 +21,8 @@
 
 abstract class EventStoreTest extends TestCase
```
asm89: Add tests for loading existing id, but different type (should throw).

asm89:
@wjzijderveld :+1: for the work on this, it had to happen eventually. :) The sooner we merge this, the better imo.

> Check if we need to normalize the classname

I suggest to do the same as we currently do for events?

> Give some hints/snippets to add a migration to your application

Yes, that'd be nice. It really depends on how you structure your own code though. I'm still thinking about this one.


> Provide a more sensible name, I don't really like stream type here

See inline comment (+1 on stream type, or just type, since we're dealing with streams already).

@jacobkiers A table per aggregate root? That's an interesting approach. These additions should perfectly fit your use case?

andersonamuller:
GetEventStore calls it `category`.

> Stream names in the Event Store are strings, and we have the convention which can be used whereby stream names of the format category-rest use the prefix before the hyphen as the event category. This can be useful later when using projections (Greg will be covering using them in his blog series on projections).

> Because the category name is used from JavaScript, the default implementation of aggregateIdToStreamName converts the name to camel case - so an aggregate of type MyTestAggregate with an ID of ffe312b9-624a-4a2a-9665-e9ae27dd1d7d ends up with the stream name myTestAggregate-ffe312b9-624a-4a2a-9665-e9ae27dd1d7d.

https://geteventstore.com/blog/20130220/getting-started-part-2-implementing-the-commondomain-repository-interface/

renan:
Couldn't it be a implementation detail? I imagine a `DBALEventStore` taking `streamType` in the constructor. Making one Event Store per Aggregate Root, like Repositories are.

francescotrucchia:
:+1:

Miliooo:
:+1:

francescotrucchia:
:+1:

Xerkus:
:-1: to your :+1:

NoelDavies:
Has any progress been made to snapshotting? Would be interested in taking a look! We use Broadway for a very specific sector in London and it works great, but going forward we can see snapshots becoming very useful in handling the number of events we're expecting.

othillo:
@NoelDavies I was looking into snapshotting myself last week. I started with a basic snapshotting repository and a snapshotting aware event store. I will submit a POC.

mhightower:
I'm not too familiar with GitHub but is the branch above the POC @othillo was talking about?

othillo:
@mhightower you can find my POC at https://github.com/othillo/broadway-snapshotting

othillo:
@wjzijderveld what is the status of this PR?

wjzijderveld:
We eventually decided not to continue with this, but decided to configure different event stores. So we never finished this integration.

I'll close this for now, if somebody wants to continue on this we can reopen, but it is already so old, that simply opening a new PR might be easier.

PR #184, opened 2015-07-08
Use require instead of require_once when retrieving loader

### LINKED ISSUE #183: Unittest require_once
Hi,

First of all this is a great project. 
Today I clone the repo to have a better look. When running the unit test an error occured.
```
PHP Fatal error:  Call to a member function add() on boolean in /var/www/broadway/test/bootstrap.php on line 14

Fatal error: Call to a member function add() on boolean in /var/www/broadway/test/bootstrap.php on line 14
```
So I looked it up and it turns out require_once returns a boolean not the content.
http://stackoverflow.com/questions/26257846/why-i-cannot-get-the-return-value-of-require-once-function-in-php

I changed require_once to require and everthing worked as expected.
I'm using php 5.6.

wjzijderveld:
Fixes #183

### DIFF
--- a/test/bootstrap.php
+++ b/test/bootstrap.php
@@ -10,7 +10,7 @@
  */
 
 if (file_exists($file = __DIR__ . '/../vendor/autoload.php')) {
-    $loader = require_once $file;
+    $loader = require $file;
     $loader->add('Broadway', __DIR__);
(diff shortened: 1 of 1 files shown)


asm89:
:+1:

PR #185, opened 2015-07-09
Solve issues with saga implementation

matthiasnoback:
In this PR I have so far fixed the following problems:

- Previously there was no way to register saga's automatically. I've added the `RegisterSagaCompilerPass` for this and you can use the `broadway.saga` service tag now.
- Saga instances should be provided as the second argument of the `MultipleSagaManager`, for which there was no service definition yet. I've added it.
- The `SagaScenarioTestCase` didn't conform to the other `*TestCase` classes in this library in that you had to create your own scenario. I changed this: it's now created for you. Since a saga is likely to use generated UUIDs, a sequence generator for them has been provided, so this allows you to provide a fixed set of UUIDs before running your test by calling `setGeneratedUuids()`. I have modified the API of the `MockUuidSequenceGenerator` to make this possible.
- Add a saga example, point to some references.
- Add a comment about the service tag.

Looking forward to your feedback!

### DIFF
--- a/examples/saga/README.md
+++ b/examples/saga/README.md
@@ -1,0 +1,53 @@
+What's a saga?
+==============
+
+A [saga](http://cqrs.nu/Faq/sagas) is a "long running" process. It allows you 
+to keep track of a (business) transaction that usually spans multiple requests.
+A saga can be seen as a [process 
+manager](http://www.enterpriseintegrationpatterns.com/ProcessManager.html), 
+which coordinates a particular message flow, possibly across bounded contexts.
+While listening to events it can decide to dispatch new commands. In order to 
+make that decision, a saga can resort to its own state.
+
+Sagas in Broadway
+-----------------
+
+In Broadway, a saga is a class that extends `Broadway\Saga\Saga`. It uses the 
+same conventions for listening to events (i.e. it has `handle*()` methods for 
+each event it's interested in). Instead of a `DomainMessage` instance, each
+of the `handle*()` methods receives a `Broadway\Saga\State` object as the
+second argument.
+
+A `handle*()` method should always return a `State` object, which will be 
+persisted afterwards. The `State` object can be used to store anything that's
+needed to retrieve the state of the saga later, or to create new commands. 
+
+Because a saga dispatches new commands, it usually accepts the command bus as 
+its first constructor argument. And since it probably needs fresh UUIDs when 
+dispatching new commands, it likely needs a UUID generator as well. The rest 
+is up to you.
+
+Sample code
+-----------
+
+The code in `ReservationSaga.php` is based on an example from the book 
+[Exploring CQRS and event sourcing (Microsoft patterns & 
+practices)[(https://msdn.microsoft.com/en-us/library/jj554200.aspx).
+
+Whenever an order is placed, a reservation is made for the requested number 
+of seats. When this reservation was accepted (i.e. the number of requested 
+seats didn't exceed the number of available seats), the order itself is marked 
+as "booked". When the reservation was rejected, the order itself is rejected.
+
+Please refer to [this 
+diagram](https://msdn.microsoft.com/en-us/library/JJ591570.20afccbda270dfd4b9cf0ffac4249b9b%28l=en-us%29.png) 
+for a full overview of the different situations that might occur.
+
+Also, take a look at the tests in `ReservationSagaTest.php` to see how you can
+prove that your saga implementation is sound.
+
+The PHPUnit tests can be run by changing to this directory and running:
+
+```bash
+phpunit .
+```
--- a/src/Broadway/Bundle/BroadwayBundle/DependencyInjection/RegisterSagaCompilerPass.php
+++ b/src/Broadway/Bundle/BroadwayBundle/DependencyInjection/RegisterSagaCompilerPass.php
@@ -1,0 +1,55 @@
+<?php
+
+namespace Broadway\Bundle\BroadwayBundle\DependencyInjection;
+
+use Symfony\Component\DependencyInjection\Compiler\CompilerPassInterface;
+use Symfony\Component\DependencyInjection\ContainerBuilder;
+use Symfony\Component\DependencyInjection\Reference;
+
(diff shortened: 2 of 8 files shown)


matthiasnoback:
@mbadolato might be interested as well :)

mbadolato:
:+1: 

Nice! I don't have time to look at the code at the moment, but yep I am interested in this! I've been able to work around needing a Saga so far, but I can feel that it's on the cusp of being part of a use case any time now :smile:

matthiasnoback:
With regard to the modified interface of `MockUuidSequenceGenerator`, I didn't notice the fact that it's inside another package. I'll create a pull request for it as well.

src/Broadway/Bundle/BroadwayBundle/DependencyInjection/BroadwayExtension.php
```diff
@@ -57,6 +57,7 @@ private function loadCommandBus(array $config, ContainerBuilder $container)
                 'broadway.command_handling.command_bus',
                 'broadway.command_handling.simple_command_bus'
             );
+            $container->removeDefinition('broadway.auditing.command_logger');
```
wjzijderveld: Why remove this :)

matthiasnoback: Ah, good question. I forgot to add a comment about it. If no logger is defined in config, the `broadway.auditing.command_logger` is still instantiated because it is tagged as an event listener. Because the logger has not been injected, it's instantiated with an empty string as the first argument, which causes a fatal error.
My solution was to remove the definition, but we could also make the service definition non-public, remove those event listener tags, or only add them if a logger has been provided.

matthiasnoback:
@wjzijderveld I removed the "remove service definition" line as discussed. I'll create a new PR to fix the issue with the command logger.

src/Broadway/Bundle/BroadwayBundle/DependencyInjection/RegisterSagaCompilerPass.php
```diff
@@ -0,0 +1,51 @@
...
+
+    public function process(ContainerBuilder $container)
+    {
+        if (!$container->has($this->multipleSagaManagerService)) {
+            throw new \LogicException(
+                sprintf('Unknown saga manager service known as %s', $this->multipleSagaManagerService)
+            );
+        }
+
+        $sagas = [];
```
sstok: Broadway is PHP 5.3 :) I'm would not have a problem with a minimum PHP version bump.

matthiasnoback:
@sstok: well spotted, fixed it.

src/Broadway/Bundle/BroadwayBundle/DependencyInjection/RegisterSagaCompilerPass.php
```diff
@@ -0,0 +1,51 @@
...
+use Symfony\Component\DependencyInjection\Compiler\CompilerPassInterface;
+use Symfony\Component\DependencyInjection\ContainerBuilder;
+use Symfony\Component\DependencyInjection\Reference;
+
+class RegisterSagaCompilerPass implements CompilerPassInterface
+{
+    private $multipleSagaManagerService;
+    private $tagName;
+
+    public function __construct($multipleSagaManagerService, $tagName)
```
wjzijderveld: Could add a docblock here?
For arguments that don't have a type hint, we like to have a docblock.

wjzijderveld:
Could you fix that 1 docblock and rebase?

matthiasnoback:
@wjzijderveld 

1. I created another PR: https://github.com/qandidate-labs/broadway/pull/189 That one should be merged first, before this PR can be merged. 
2. The same goes for https://github.com/qandidate-labs/broadway-uuid-generator/issues/2 That one should be merged and released first, and then we need to update this package itself to use the latest uuid-generator version.
3. I'll add the docblock you requested.

matthiasnoback:
@wjzijderveld Just added the docblock.

matthiasnoback:
@wjzijderveld Instead of creating the UUID generator first, and later setting the UUIDs, I now leave that to the user in the [`createSaga()` method](https://github.com/qandidate-labs/broadway/pull/185/files). This makes the most sense to me in the end. I'll close the PR on the uuid-generator package as well.
(Why does this make sense to me: having a `setUuids()` method on the UUID generator is a recipe for bugs because it makes it unpredictable. Also, doing it like this forces the user to think about what they need as constructor arguments. Also, not every saga needs a uuid generator I guess.)

matthiasnoback:
Rebased on the latest master.

matthiasnoback:
No more TODOs :) What do you think of the example?

wjzijderveld:
I need a bit more time to look at this PR, will come back asap. :smile:

matthiasnoback:
No problem, if you like I can also put the example in a separate PR.
Op 15 jul. 2015 09:47 schreef "Willem-Jan Zijderveld" <
notifications@github.com>:

matthiasnoback:
Rebased on the latest master.

wjzijderveld:
Looks good to me!

@qandidate-labs/broadway Can somebody else look at this as well?

examples/saga/README.md
```diff
@@ -0,0 +1,53 @@
...
+its first constructor argument. And since it probably needs fresh UUIDs when 
+dispatching new commands, it likely needs a UUID generator as well. The rest 
+is up to you.
+
+Sample code
+-----------
+
+The code in `ReservationSaga.php` is based on an example from the book 
+[Exploring CQRS and event sourcing (Microsoft patterns & 
+practices)[(https://msdn.microsoft.com/en-us/library/jj554200.aspx).
```
asm89: nit: [ too many on this line, making the markdown not render properly

examples/saga/README.md
```diff
@@ -0,0 +1,53 @@
+What's a saga?
```
asm89: :+1:

asm89:
:+1: :shipit:

wjzijderveld:
Thanks @matthiasnoback!

asm89:
:tada:

PR #188, opened 2015-07-13
Allow for event store management; query event store.

simensen:
This is an alternate implementation to #73. It follows some of the original AxonFramework naming and ideas but does so with a much simpler interface for building and using criteria. Essentially, I'm allowing for filtering/querying just on Aggregate Root Type, Aggregate Root ID, and Event Type.

There is currently no way to actually filter/query based on Aggregate Root Type but I didn't realize that until after it was completely implemented. :) I thought I'd be able to do it for at least the in memory store but nope! I can remove it and put it back in after #181.

```php
$criteria = Criteria::create()->withEventType('Acme.Foo.Event.EatFoo');

$criteria = Criteria::create()->withEventTypes([
    'Acme.Foo.Event.EatFoo',
    'Acme.Foo.Event.CookFoo',
]);

$criteria = Criteria::create()->withEventType('Acme.Foo.Event.EatFoo');

$criteria = Criteria::create()
    ->withAggregateRootId('CA59B2DB-7782-4A63-9304-91DDEBB425E3')
    ->withEventType('Acme.Foo.Event.EatFoo')
;

$eventStore->visitEvents($eventVisitor, $criteria);
```

There are probably some PHP 5.4 `[]` arrays somewhere. Will see once Travis runs. :)

Tests are written for In Memory, DBAL, and DBAL w/ Binary.

### DIFF
--- a/src/Broadway/EventStore/CallableEventVisitor.php
+++ b/src/Broadway/EventStore/CallableEventVisitor.php
@@ -11,7 +11,6 @@
 
 namespace Broadway\EventStore;
 
-
 use Broadway\Domain\DomainMessage;
 
 class CallableEventVisitor implements EventVisitorInterface
--- a/src/Broadway/EventStore/CallableEventVisitor.php
+++ b/src/Broadway/EventStore/CallableEventVisitor.php
@@ -1,0 +1,33 @@
+<?php
+
+/*
+ * This file is part of the broadway/broadway package.
+ *
+ * (c) Qandidate.com <opensource@qandidate.com>
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+namespace Broadway\EventStore;
+
+
+use Broadway\Domain\DomainMessage;
+
+class CallableEventVisitor implements EventVisitorInterface
+{
+    /**
+     * @var callable
+     */
+    private $callable;
+
+    public function __construct(callable $callable)
+    {
+        $this->callable = $callable;
+    }
+
+    public function doWithEvent(DomainMessage $domainMessage)
+    {
+        call_user_func($this->callable, $domainMessage);
+    }
+}
--- a/src/Broadway/EventStore/DBALEventStore.php
+++ b/src/Broadway/EventStore/DBALEventStore.php
@@ -282,9 +282,10 @@ class DBALEventStore implements EventStoreInterface, EventStoreManagementInterfa
             $criteriaTypes[] = array('uuid IN (:uuids)');
 
             if ($this->useBinary) {
-                $bindValues['uuids'] = array_map(function ($id) {
-                    return $this->convertIdentifierToStorageValue($id);
-                }, $criteria->getAggregateRootIds());
+                $bindValues['uuids'] = array();
+                foreach ($criteria->getAggregateRootIds() as $id) {
+                    $bindValues['uuids'][] = $this->convertIdentifierToStorageValue($id);
+                }
                 $bindValueTypes['uuids'] = Connection::PARAM_STR_ARRAY;
             } else {
                 $bindValues['uuids'] = $criteria->getAggregateRootIds();
--- a/src/Broadway/EventStore/DBALEventStore.php
+++ b/src/Broadway/EventStore/DBALEventStore.php
@@ -24,7 +24,6 @@ use Doctrine\DBAL\Connection;
 use Doctrine\DBAL\DBALException;
 use Doctrine\DBAL\Schema\Schema;
 use Doctrine\DBAL\Version;
-use Elasticsearch\Tests\Endpoint\Indicess\CreateTest;
 use Rhumsaa\Uuid\Uuid;
 
 /**
@@ -81,9 +80,6 @@ class DBALEventStore implements EventStoreInterface, EventStoreManagementInterfa
 
         $events = array();
         while ($row = $statement->fetch()) {
-            if ($this->useBinary) {
-                $row['uuid'] = $this->convertStorageValueToIdentifier($row['uuid']);
-            }
             $events[] = $this->deserializeEvent($row);
         }
 
@@ -198,7 +194,7 @@ class DBALEventStore implements EventStoreInterface, EventStoreManagementInterfa
     private function deserializeEvent($row)
     {
         return new DomainMessage(
-            $row['uuid'],
+            $this->convertStorageValueToIdentifier($row['uuid']),
             $row['playhead'],
             $this->metadataSerializer->deserialize(json_decode($row['metadata'], true)),
             $this->payloadSerializer->deserialize(json_decode($row['payload'], true)),
@@ -236,16 +232,12 @@ class DBALEventStore implements EventStoreInterface, EventStoreManagementInterfa
         return $id;
     }
 
-    public function visitEvents(EventVisitorInterface $eventVisitor, Criteria $criteria = null)
+    public function visitEvents(Criteria $criteria, EventVisitorInterface $eventVisitor)
     {
-        $statement = $this->prepareVisitEventsStatement($criteria ?: Criteria::create());
+        $statement = $this->prepareVisitEventsStatement($criteria);
         $statement->execute();
 
         while ($row = $statement->fetch()) {
-            if ($this->useBinary) {
-                $row['uuid'] = $this->convertStorageValueToIdentifier($row['uuid']);
-            }
-
             $domainMessage = $this->deserializeEvent($row);
 
             $eventVisitor->doWithEvent($domainMessage);
@@ -279,7 +271,7 @@ class DBALEventStore implements EventStoreInterface, EventStoreManagementInterfa
         $criteriaTypes = array();
 
         if ($criteria->getAggregateRootIds()) {
-            $criteriaTypes[] = array('uuid IN (:uuids)');
+            $criteriaTypes[] = 'uuid IN (:uuids)';
 
             if ($this->useBinary) {
                 $bindValues['uuids'] = array();
@@ -294,7 +286,7 @@ class DBALEventStore implements EventStoreInterface, EventStoreManagementInterfa
         }
 
         if ($criteria->getEventTypes()) {
-            $criteriaTypes[] = array('type IN (:types)');
+            $criteriaTypes[] = 'type IN (:types)';
             $bindValues['types'] = $criteria->getEventTypes();
             $bindValueTypes['types'] = Connection::PARAM_STR_ARRAY;
         }
@@ -303,11 +295,7 @@ class DBALEventStore implements EventStoreInterface, EventStoreManagementInterfa
             return array('', array(), array());
         }
 
-        $where = 'WHERE '.join(' AND ', array_map(function (array $criteriaType) {
-            return implode(' OR ', $criteriaType);
-        }, array_filter($criteriaTypes, function ($input) {
-            return $input;
-        })));
+        $where = 'WHERE '.join(' AND ', $criteriaTypes);
 
         return array($where, $bindValues, $bindValueTypes);
     }
--- a/src/Broadway/EventStore/DBALEventStore.php
+++ b/src/Broadway/EventStore/DBALEventStore.php
@@ -16,11 +16,15 @@ use Broadway\Domain\DomainEventStream;
 use Broadway\Domain\DomainEventStreamInterface;
 use Broadway\Domain\DomainMessage;
 use Broadway\EventStore\Exception\InvalidIdentifierException;
+use Broadway\EventStore\Management\Criteria;
+use Broadway\EventStore\Management\CriteriaNotSupportedException;
+use Broadway\EventStore\Management\EventStoreManagementInterface;
 use Broadway\Serializer\SerializerInterface;
 use Doctrine\DBAL\Connection;
 use Doctrine\DBAL\DBALException;
 use Doctrine\DBAL\Schema\Schema;
 use Doctrine\DBAL\Version;
+use Elasticsearch\Tests\Endpoint\Indicess\CreateTest;
 use Rhumsaa\Uuid\Uuid;
 
 /**
@@ -29,7 +33,7 @@ use Rhumsaa\Uuid\Uuid;
  * The implementation uses doctrine DBAL for the communication with the
  * underlying data store.
  */
-class DBALEventStore implements EventStoreInterface
+class DBALEventStore implements EventStoreInterface, EventStoreManagementInterface
 {
     private $connection;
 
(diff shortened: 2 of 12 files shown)


simensen:
One thing I noticed while working on this is that DBAL implementation is that there is a sort by playhead. Is it not safe to assume that the id (autoincrement?) is the correct order for the events? Under what circumstances might the playhead sequence be out of order w/ the actual event id sequence?

wjzijderveld:
The playhead is determined as soon as the event is recorded (when the DomainMessage is created). So the software determines in what order events occurred. Especially when you are going to work with async, it's possible that 2 threads are going to use the same playhead, causing an unique constraint error.
When only using the auto-increment, MySQL would just insert both events.

simensen:
So is it safe to rely on the auto increment ID for event order?

matthiasnoback:
This sounds very useful, @simensen. Yesterday I was thinking, because of @wjzijderveld's post on replaying events: 

- You can programmatically find out which projectors exist.
- By looping through their `apply*()` methods you can also find out which events are of interest to them.
- By convention, their first constructor argument is the view model repository.

So if I'm correct, you can fully automate the process of recreating each type of view model separately. If the projector is in every way idempotent. If it isn't, you may need to implemet a `removeAll()` method for view model repositories.

Do you think this might work?

simensen:
@matthiasnoback I think that is an interesting idea. Because the tools haven't really existed for Broadway yet I haven't seen a working end-to-end "destroy and rebuild" workflow. The conventions stuff is interesting and Broadway definitely ships with a bunch of convention-based stuff out of the box so building something like that (and hooking it into the Symfony bundle) could be quite interesting.

One things to keep in mind is that sometimes you might just want to rebuild one projection and not all of the projections in the system. So I'm not sure how useful "rebuilding all of the projections" would be but I'm sure for smaller projects that might be a nice thing to do from time to time?

If this was done in Broadway itself, it might be nice to have this removeAll being part of an additional interface that a read model can implement rather than requiring all read models implement it.

wjzijderveld:
@matthiasnoback It should be possible to create something like this for your application. I highly doubt this is something Broadway will have out of the box on the short term. Especially all ties to conventions holds me back :smile: 

That we have quit a lot of convention based stuff now, is mainly because it was developed inside a project and it made sense for us at the time. 

But that's just me :smile:  , maybe others from @qandidate-labs/broadway have another opinion?

matthiasnoback:
@wjzijderveld Totally agree, the solution may be generic, but it would only work if you follow the conventions everywhere. I don't think it belongs in the project repository as well. If I ever try this, it will be a separate library (by the way, having a `removeAll()` method on an interface somewhere, would also help us to encapsulate this :) https://github.com/qandidate-labs/broadway-demo/blob/master/test/BroadwayDemoBundle/WebTestCase.php#L74).

src/Broadway/EventStore/DBALEventStore.php
```diff
@@ -16,11 +16,15 @@
...
 use Broadway\EventStore\Exception\InvalidIdentifierException;
+use Broadway\EventStore\Management\Criteria;
+use Broadway\EventStore\Management\CriteriaNotSupportedException;
+use Broadway\EventStore\Management\EventStoreManagementInterface;
 use Broadway\Serializer\SerializerInterface;
 use Doctrine\DBAL\Connection;
 use Doctrine\DBAL\DBALException;
 use Doctrine\DBAL\Schema\Schema;
 use Doctrine\DBAL\Version;
+use Elasticsearch\Tests\Endpoint\Indicess\CreateTest;
```
wjzijderveld: This class is not used

src/Broadway/EventStore/Management/Criteria.php
```diff
@@ -0,0 +1,132 @@
...
+
+    public function withAggregateRootTypes(array $aggregateRootTypes)
+    {
+        $instance = clone($this);
+        $instance->aggregateRootTypes = $aggregateRootTypes;
+
+        return $instance;
+    }
+
+    public function withAggregateRootType($aggregateRootType)
```
wjzijderveld: Could you add a docblock? `@param string $aggregateRootType`

wjzijderveld: As well as the other methods that don't have typehints in this class :)

wjzijderveld:
Some minor CS comments, otherwise looks good to me.

@qandidate-labs/broadway Could somebody else take a look as well?

src/Broadway/EventStore/InMemoryEventStore.php
```diff
@@ -64,4 +68,16 @@ private function assertPlayhead($events, $playhead)
             );
         }
     }
+
+    public function visitEvents(EventVisitorInterface $eventVisitor, Criteria $criteria = null)
+    {
+        $allEvents = $this->allEvents;
+        if (! is_null($criteria)) {
+            $allEvents = array_filter($allEvents, array($criteria, 'isMatchedBy'));
```
asm89: nit: I'd move the filtering to the foreach

src/Broadway/EventStore/CallableEventVisitor.php
```diff
@@ -0,0 +1,33 @@
...
+ *
+ * (c) Qandidate.com <opensource@qandidate.com>
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+namespace Broadway\EventStore;
+
+
```
asm89: nit: whitespace

src/Broadway/EventStore/Management/EventStoreManagementInterface.php
```diff
@@ -0,0 +1,19 @@
...
+ * file that was distributed with this source code.
+ */
+
+namespace Broadway\EventStore\Management;
+
+use Broadway\EventStore\EventVisitorInterface;
+
+interface EventStoreManagementInterface
+{
+    public function visitEvents(EventVisitorInterface $eventVisitor, Criteria $criteria = null);
```
asm89: I'd rather not allow `Criteria` to be `null`. People can explicitly pass an empty criteria object if they need it.

If you agree, let's also swap the arguments.

src/Broadway/EventStore/Management/CriteriaNotSupportedException.php
```diff
@@ -0,0 +1,26 @@
...
+ * Criteria not supported by implementation
+ *
+ * In some cases an event store implementation may implement management
+ * but not be able to satisfy all criteria options. In this case, the
+ * implementation must throw this exception.
+ *
+ * Class CriteriaNotSupportedException
+ * @package Broadway\EventStore\Management
+ */
+class CriteriaNotSupportedException extends EventStoreManagementException
```
asm89: :+1:

src/Broadway/EventStore/DBALEventStore.php
```diff
@@ -231,4 +235,80 @@ private function convertStorageValueToIdentifier($id)
...
+        }
+    }
+
+    private function prepareVisitEventsStatement(Criteria $criteria)
+    {
+        list ($where, $bindValues, $bindValueTypes) = $this->prepareVisitEventsStatementWhereAndBindValues($criteria);
+        $query = 'SELECT uuid, playhead, metadata, payload, recorded_on
+            FROM ' . $this->tableName . '
+            ' . $where . '
+            ORDER BY id ASC';
```
asm89: Can you move the logic to go from `Criteria` to its own class? Something like:
- `DBALCriteriaQuery($query, $bindValues, $bindValuesTypes);`
- `DBALCriteriaQueryBuilder`

I guess instead of having a builder class it could also be a static method on the `DBALCriteriaQuery`.

src/Broadway/EventStore/DBALEventStore.php
```diff
@@ -231,4 +235,80 @@ private function convertStorageValueToIdentifier($id)
...
         return $id;
     }
+
+    public function visitEvents(EventVisitorInterface $eventVisitor, Criteria $criteria = null)
+    {
+        $statement = $this->prepareVisitEventsStatement($criteria ?: Criteria::create());
+        $statement->execute();
+
+        while ($row = $statement->fetch()) {
+            if ($this->useBinary) {
```
asm89: Should we push this down to `deserializeEvent()`?

src/Broadway/EventStore/DBALEventStore.php
```diff
@@ -231,4 +235,80 @@ private function convertStorageValueToIdentifier($id)
...
+            $bindValueTypes['types'] = Connection::PARAM_STR_ARRAY;
+        }
+
+        if (! $criteriaTypes) {
+            return array('', array(), array());
+        }
+
+        $where = 'WHERE '.join(' AND ', array_map(function (array $criteriaType) {
+            return implode(' OR ', $criteriaType);
+        }, array_filter($criteriaTypes, function ($input) {
```
asm89: Why this `array_filter`? `$criteriaTypes` should not contain any `null` values at this point.

src/Broadway/EventStore/DBALEventStore.php
```diff
@@ -231,4 +235,80 @@ private function convertStorageValueToIdentifier($id)
...
+            $bindValues['types'] = $criteria->getEventTypes();
+            $bindValueTypes['types'] = Connection::PARAM_STR_ARRAY;
+        }
+
+        if (! $criteriaTypes) {
+            return array('', array(), array());
+        }
+
+        $where = 'WHERE '.join(' AND ', array_map(function (array $criteriaType) {
+            return implode(' OR ', $criteriaType);
```
asm89: In the code above `$criteriaType` will always contain only 1 value. Would it make sense to just make it a `string` instead of an `array<string>` and get rid of this `array_map()`?

src/Broadway/EventStore/InMemoryEventStore.php
```diff
@@ -53,6 +56,7 @@ public function append($id, DomainEventStreamInterface $eventStream)
             $this->assertPlayhead($this->events[$id], $playhead);
 
             $this->events[$id][$playhead] = $event;
+            $this->allEvents[] = $event;
```
asm89: Why this addition? We can also just loop over `events[]`.

src/Broadway/EventStore/Management/Criteria.php
```diff
@@ -0,0 +1,132 @@
...
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+namespace Broadway\EventStore\Management;
+
+use Broadway\Domain\DomainMessage;
+
+class Criteria
```
asm89: @simensen I'm a bit conflicted about the amount of `with*()` methods here.

test/Broadway/EventStore/Management/EventStoreManagementTest.php
```diff
@@ -0,0 +1,348 @@
...
+            $this->createDomainMessage(3, 5, new End()),
+        ), $visitedEvents);
+    }
+
+    /** @test */
+    public function it_visits_additional_aggregate_root_ids()
+    {
+        $visitedEvents = $this->visitEvents(Criteria::create()
+            ->withAggregateRootId($this->getId(1))
+            ->withAdditionalAggregateRootId($this->getId(3))
```
asm89: this is mostly testing the `Criteria` DSL. I'd say a test for 1 and multiple aggregate root ids is fine.

asm89:
@simensen Great work on this code. Not all my comments are set in stone, happy to discuss. Thank you for your work on it!

src/Broadway/EventStore/InMemoryEventStore.php
simensen: I don't think you'd be able to get all events in order that way. You'd get all events in order *per aggregate root*. So if I wanted all events from the beginning of time, I'd get them all in clusters of individual aggregate roots. By adding `allEvents` we can get them in order that they were actually added.

Other two options:

 * Do some sort of deep sorting thing based on the time the events were added.
 * Rewrite the code that is aggregate root specific to read through `allEvents` each time rather than having them hashed like this.

Also possible I'm completely missing something obvious. :) But this was the rationale for adding this extra property.

src/Broadway/EventStore/Management/Criteria.php
simensen: I could remove the singular withs. Would that help?

src/Broadway/EventStore/InMemoryEventStore.php
asm89: ### Ordering guarantees

Ah! Check. So one thing we haven't discussed so far is the guarantees we want to provide in `visitEvents()`. I think we should:

1. _guarantee_ events per aggregate are replayed in the order they were saved
2. _encourage_ chronical ordering

For the SQL world providing both (1) and (2) makes sense, but imagine an event store using a document based model. One approach there would be to store one document per id, containing all the events. While guaranteeing (1) is easy to do, guaranteeing (2) will be hard.

### In Memory

In case of the in memory implementation it is (probably?) only used in tests. Most tests focus on one aggregate root id anyway. While we encourage total chronological ordering, I think it might actually be a good idea to not provide that ordering in the in memory event store, so that people will be made aware of the guarantees we make.

src/Broadway/EventStore/Management/Criteria.php
asm89: Yes, let's start more minimal. :+1:

src/Broadway/EventStore/InMemoryEventStore.php
simensen: I thought about this more and am realizing it would make it difficult to build a common test case for this if you cannot rely on different implementations returning the events in the same order.

I think that in general the idea here is sound, but I'm also thinking this could be problematic if you have a projector that relies on events from multiple aggregate root types. You may run into a case where you won't be able to rebuild the read model in the same way you expect.

I don't know if there is a way around this, though, so it may just have to be something that we have to accept. :)

asm89: > I thought about this more and am realizing it would make it difficult to build a common test case for this if you cannot rely on different implementations returning the events in the same order.

A common test case is possible, but the assertions are a bit more complicated. The assertion would be:
- the _set_ of events returned when visiting is equal to the expected _set_
- the order of the events _per aggregate_ are ordered

I'd be happy to chat on IRC to help you generalize the test cases like this. Shouldn't be too hard.

> I think that in general the idea here is sound, but I'm also thinking this could be problematic if you have a projector that relies on events from multiple aggregate root types. You may run into a case where you won't be able to rebuild the read model in the same way you expect.

Projectors that rely on events from multiple aggregate root types are tricky in any case. If a specific implementation does guarantee to return events in chronological order that's great of course and as a developer you can leverage that fact. It's just hard to guarantee in this library in the general case.

In light of the discussion above I'm conflicted about what to do with the `InMemoryEventStore`. Perhaps it should be configurable? In any case I think we should ship the non-chronological ordered `InMemoryEventStore` in this PR and possibly follow-up with extra things if folks are interested.

src/Broadway/EventStore/DBALEventStore.php
simensen: Weird, I don't remember ever adding that. :)

simensen:
I removed a whole lot of `with*` methods from `Criteria`, made `Criteria` required and non-nullable, swapped the argument order, removed `$allEvents` from `InMemoryEventStore`, made the checks based on checking order of aggregate type and aggregate ID, and fixed most of the other CS issues.

The only thing I didn't do was extract the criteria query building stuff from `DBALEventStore`. I looked at it and realized that doing so would probably require either 1) more refactoring than I have time for now or 2) code duplication. I don't think I'll have time to do either of these right now.

The biggest issue was the dependency on `isBinary` for how the bind values and bind types are created. In addition to that, we'd also need access to `convertIdentifierToStorageValue`.

My vote is to merge it as-is. :) However, if someone else (@asm89 offered?) wants to finish this up I'm all for it. :)

src/Broadway/EventStore/DBALEventStore.php
simensen: See the comment I just added on this. `prepareVisitEventsStatementWhereAndBindValues` requires access to `isBinary` and `convertIdentifierToStorageValue` which means either refactoring this class in a larger way, exposing those variables via public methods, or duplicating the code in `DBALCriteriaQuery`. I don't think I have the time to do any of these right now (though public methods would probably be pretty trivial) so if this is super important this is probably a good candidate for someone else taking over to submit as a PR against my PR before we merge it.

asm89:
@simensen +1 for merging as is. Can you squash your commits?

asm89:
Shipped it in https://github.com/qandidate-labs/broadway/commit/e81d4ea167ce97383a9a4b7d85542e8b5e11900a. Thank you for your work on this!

PR #191, opened 2015-07-17
Make the `broadway:event-store:schema:drop` idempotent.

kimlai:
When this command is run, it should not throw an exception is the table does not exist. This is useful for CI scripting.


wjzijderveld:
Thanks @kimlai 
Will merge when Travis decides to work :)

othillo:
:+1:

wjzijderveld:
@kimlai Could you rebase this on master? Composer kept failing on the old Travis infrastructure, so upgraded it to use the container platform. That seems to work :smile:

kimlai:
It seems to work indeed :)

wjzijderveld:
Thanks @kimlai !

PR #192, opened 2015-07-18
Lets see if Composer works on Travis Container Platform

wjzijderveld:


Xerkus:
It does work in containers

PR #193, opened 2015-07-22
Fix DBAL Event Store to work with mysqli extension

jacobkiers:
The Mysqli extension does not handle named parameters in a prepared
statement. Therefore, it needs to be positional.


jacobkiers:
Seems the build failed due to composer errors. Could someone restart it?

sstok:
Amend the last commit and then force push to trigger a new build.

jacobkiers:
Didn't know of that trick. Will use it for sure next time. Thanks!

asm89:
lgtm :+1:

wjzijderveld:
Thanks @jacobkiers

PR #198, opened 2015-07-27
Bugfix: Avoid tagging the saga manager when no sagas are registered

### LINKED ISSUE #197: Doctrine MongoDB is now a requirement
After getting the new version 0.7, doctrine mongodb seems to be a requirement:

Fatal error: Class 'Doctrine\MongoDB\Connection' not found in /data/app/cache/dev/appDevDebugProjectContainer.php on line 644

If this was intentional, doctrine-mongodb should be moved from require-dev to require in composer. Otherwise, it should be handled correctly in the DI container.

wjzijderveld:
Fixes #197 

As the `MultipleSagaManager` was registered to the event bus, the mongo connection turned into a hard requirement.
This PR fixes that requirement by only tagging the manager if there are actual sagas registered.

### DIFF
--- a/src/Broadway/Bundle/BroadwayBundle/BroadwayBundle.php
+++ b/src/Broadway/Bundle/BroadwayBundle/BroadwayBundle.php
@@ -35,6 +35,12 @@ class BroadwayBundle extends Bundle
         parent::build($container);
 
         $container->addCompilerPass(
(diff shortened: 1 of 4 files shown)


asm89:
:+1:

matthiasnoback:
Good catch! :)

PR #204, opened 2015-08-14
IlluminateEventStore

dsgepers:
IlluminateEventStore for use with Laravel and Lumen Frameworks.

This uses Illuminates query builder instead of Eloquent, since Eloquent would require the models, and Lumen's default configuration does not have Eloquent enabled.

### DIFF
--- a/src/Broadway/EventStore/IlluminateEventStore.php
+++ b/src/Broadway/EventStore/IlluminateEventStore.php
@@ -6,9 +6,9 @@ use Broadway\Domain\DomainMessage;
 use Broadway\Domain\DateTime;
 use Broadway\EventStore\EventStoreInterface;
 use Broadway\EventStore\EventStreamNotFoundException;
-use Broadway\EventStore\IlluminateEventStoreException;
 use Broadway\Serializer\SerializerInterface;
-use Illuminate\Contracts\Container\Container;
+use Illuminate\Config\Repository;
+use Illuminate\Database\DatabaseManager;
 
 /**
  * Class IlluminateEventStore
@@ -21,7 +21,7 @@ use Illuminate\Contracts\Container\Container;
 class IlluminateEventStore implements EventStoreInterface {
 
     /**
-     * @var \Illuminate\Database\DatabaseManager
+     * @var DatabaseManager
      */
     protected $databaseManager;
 
@@ -38,15 +38,17 @@ class IlluminateEventStore implements EventStoreInterface {
     /**
      * Construct the dependancies
      *
-     * @param \Illuminate\Contracts\Container\Container $container
+     * @param \Illuminate\Database\DatabaseManager $databaseManager
+     * @param \Illuminate\Config\Repository $configRepository
      * @param SerializerInterface $serializer
      */
     public function __construct(
-        Container $container,
+        DatabaseManager $databaseManager,
+        Repository $configRepository,
         SerializerInterface $serializer
     ) {
-        $this->databaseManager = $container->make('db');
-        $this->config = $container->make('config');
+        $this->databaseManager = $databaseManager;
+        $this->config = $configRepository;
         $this->table = $this->config->get('broadway.table','event');
         $this->serializer = $serializer;
     }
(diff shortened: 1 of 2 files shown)


src/Broadway/EventStore/IlluminateEventStore.php
```diff
@@ -0,0 +1,117 @@
...
+    protected $config;
+
+    /**
+     * Construct the dependancies
+     *
+     * @param \Illuminate\Contracts\Container\Container $container
+     * @param SerializerInterface $serializer
+     */
+    public function __construct(
+        Container $container,
```
pminnieur: Why not inject the concrete instances instead of a Container?

dsgepers: Yea, that seems right.

asm89:
@dsgepers Thank you for your work on this! I'm wondering if it would make more sense to move this to the laravel broadway integration though? https://github.com/nWidart/Laravel-broadway

cc @nWidart @MarkRedeman

src/Broadway/EventStore/IlluminateEventStore.php
```diff
@@ -0,0 +1,119 @@
...
+     * @param SerializerInterface $serializer
+     */
+    public function __construct(
+        DatabaseManager $databaseManager,
+        Repository $configRepository,
+        SerializerInterface $serializer
+    ) {
+        $this->databaseManager = $databaseManager;
+        $this->config = $configRepository;
+        $this->table = $this->config->get('broadway.table','event');
```
asm89: I don't know enough about Laravel and friends, but wouldn't it be possible to inject:
- the database connection
- the table

Instead of the database manager and the configuration?

dsgepers: While I've seen the Laravel broadway implementation, I acctualy prefered to implement broadway plain-and-straight. My main project uses Lumen, and the package doesn't mention anything about it.

Concerning the injections:
* While it would be possible to directly inject the Connection, it would require the developer to configure an additional binding within it's application. The current solution would work out of the box.
* The table name can be set with laravels config object, (but it's not a requirement as it defaults to 'event').

The choices I made were basicaly done with the mind set of doing minimal additional configuration when using this implementation.

nWidart:
Hello @dsgepers ,

The laravel-broadway package is not yet supporting Lumen, but that doesn't mean it will never. I'm not using Lumen at all, so maybe someone can send a PR to add support for that ?

About adding this to the laravel package or Broadway itself, I'm not sure what the most ideal would be.

On one hand it makes sense to have it inside the Laravel-Broadway, since non laravel users don't need the Illuminate implementation.
Though this is not necessarily true for the future, since Eloquent could be used inside Symfony for instance.

On the other hand there's the Dbal event store that's in this repository, so it would make sense to add this to the current repository.

It's something that not all users will require/need, but, it will also not slow them down if they don't.

MarkRedeman:
@nWidart I'm currently working on a hobby project that uses Lumen and will probably use Broadway. So if no one else has made a PR by the time that I'm working on that integration then I probably will.

---

Perhaps we could split up the eventstore implementations into separate packages.

sstok:
The Symfony bundle was added because it was easier for the developers, but according to https://leanpub.com/principles-of-package-design/ (must read) these kind of additions should life in there own repository as they have there own dependencies that will not work otherwise.
http://php-and-symfony.matthiasnoback.nl/2014/04/theres-no-such-thing-as-an-optional-dependency/

So even this is accepted it should live in its own repository, same goes for the Symfony bundle ;)

cordoval:
:+1: great idea @sstok

mbadolato:
IIRC, the plan is to move the Symfony bundle to its own repository, it was just [someone, can't remember who] preference to leave it here for ease of management until a stable (1.0?) version was reached

wjzijderveld:
Correct @mbadolato, we also still want to split the packages itself sometime, but not yet :smile:

othillo:
I'm closing this for now. Additional event stores are very welcome, though! We are actually splitting implementations to separate repositories at the moment. If someone wants to work on a new broadway/event-store-illuminate, please ping us! /cc @broadway/broadway

PR #205, opened 2015-08-18
Factory definitions in extension

qooplmao:
Moved the factory definitions to the extension so that they can be set correctly before and after the change in `symfony/dependency-injection` 2.6.


wjzijderveld:
This is the solution I was thinking about as well, don't really like that we have to do it, but I guess we don't have much choice :disappointed: 

ping @qandidate-labs/broadway

wjzijderveld:
Fixed with #228 
We dropped support for DependencyInjection <2.6

PR #206, opened 2015-08-19
Change DateTime to be locale agnostic.

bburnichon:
`sprintf('%.6f', $float)` is locale aware. When trying to convert a float to a server with locale set to `'fr_FR'` for example, the result string has `,` inside as a decimal separator.

This leads to `DateTime::createFromFormat()` to fail as it expects `.` instead.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -17,6 +17,12 @@ matrix:
           env: COMPOSER_LOWEST="--prefer-lowest"
     fast_finish: true
 
+addons:
+    apt:
+        packages:
+            - language-pack-fr
+            - language-pack-en
+
 before_install:
     - echo "extension=mongo.so" >> ~/.phpenv/versions/$(phpenv version-name)/etc/php.ini
     - echo "memory_limit=2G" >> ~/.phpenv/versions/$(phpenv version-name)/etc/conf.d/travis.ini
--- a/test/Broadway/Domain/DateTimeTest.php
+++ b/test/Broadway/Domain/DateTimeTest.php
@@ -40,8 +40,8 @@ class DateTimeTest extends TestCase
     public function providesLocales()
     {
         return array(
-            array('en_GB'),
-            array('fr_FR'),
+            array('en_GB.UTF-8'),
+            array('fr_FR.UTF-8'),
         );
     }
 
--- a/test/Broadway/Domain/DateTimeTest.php
+++ b/test/Broadway/Domain/DateTimeTest.php
@@ -28,10 +28,21 @@ class DateTimeTest extends TestCase
 
     /**
      * @test
+     * @dataProvider providesLocales
      */
-    public function it_creates_now()
+    public function it_creates_now($locale)
     {
+        setlocale(LC_ALL, $locale);
+
         $this->assertInstanceOf('Broadway\Domain\DateTime', DateTime::now());
+    }
+
+    public function providesLocales()
+    {
+        return array(
+            array('en_GB'),
+            array('fr_FR'),
+        );
     }
 
     /**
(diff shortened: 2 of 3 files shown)


sstok:
:+1:

test/Broadway/Domain/DateTimeTest.php
```diff
@@ -28,12 +28,23 @@ public function it_converts_back_and_forth()
 
     /**
      * @test
+     * @dataProvider providesLocales
      */
-    public function it_creates_now()
+    public function it_creates_now($locale)
     {
+        setlocale(LC_ALL, $locale);
```
Maks3w: use `$this->setLocale()` instead.

>  This method is a wrapper for the setlocale() function that automatically
>   resets the locale to its original value after the test is run.

bburnichon: Did not know about this `setLocale()` method. PR updated.

wjzijderveld:
:+1: from me when changing the `$this->setLocale()` (I didn't knew that existed :smile:)

test/Broadway/Domain/DateTimeTest.php
```diff
@@ -28,12 +28,23 @@ public function it_converts_back_and_forth()
 
     /**
      * @test
+     * @dataProvider providesLocales
      */
-    public function it_creates_now()
+    public function it_creates_now($locale)
     {
+        $this->setLocale(LC_ALL, $locale);
```
wjzijderveld: Don't think it matters that much, but we could use LC_TIME here.
But as it resets anyway, I'm fine with using LC_ALL here.

wjzijderveld:
Ah yes, Travis and locales... 
I think Travis by default only has en_US installed, so en_GB and fr_FR don't work.
See: http://docs.travis-ci.com/user/common-build-problems/#System:-Required-language-pack-isn’t-installed

test/Broadway/Domain/DateTimeTest.php
bburnichon: It's actually not LC_TIME that causes issue here, but LC_NUMERIC (float number format). I think it's best to keep LC_ALL.

Maks3w:
Note: For Travis containers this is the way for to do APT installs

http://docs.travis-ci.com/user/migrating-from-legacy/#How-do-I-install-APT-sources-and-packages%3F

test/Broadway/Domain/DateTimeTest.php
```diff
@@ -28,12 +28,23 @@ public function it_converts_back_and_forth()
...
+
         $this->assertInstanceOf('Broadway\Domain\DateTime', DateTime::now());
     }
 
+    public function providesLocales()
+    {
+        return array(
+            array('en_GB'),
```
Maks3w: Try to append `.UTF8` at the end of the locale name

.travis.yml
```diff
@@ -17,7 +17,15 @@ matrix:
           env: COMPOSER_LOWEST="--prefer-lowest"
     fast_finish: true
 
+addons:
+    apt:
+        packages:
+            - language-pack-fr
+            - language-pack-en
+            - language-pack-de
```
Maks3w: Why to install DE if there is no test for it

bburnichon: An error while copying install procedure from Travis doc. I did not remembered which locale were already used.

DE locale removed in last amending commit

test/Broadway/Domain/DateTimeTest.php
bburnichon: Thanks, did not noticed that was the issue with tests.

bburnichon:
Now green on travis :smile:

Maks3w:
And most important. Before the test was wrongly green.

sstok:
What is the status of this pr?

PR #207, opened 2015-08-21
Feature in memory repository data protected

vladylen:


wjzijderveld:
Why do you need this change?

vladylen:
sorry for long response...
I've inherited this class and changed it for working with auto increment id (maybe I should not use auto increment id at all, but ...), with 'protected' I should not duplicate methods that are valid in this case.
If it's contradict with your private/protected agreement, just cancel this PR

othillo:
@vladylen I'm closing this one.

PR #209, opened 2015-09-06
[Travis] multiple improvements

### LINKED ISSUE #76: Add HHVM to travis build matrix
It is a bit more work then just adding it to the build matrix.

The mongo extension isn't available yet for hhvm, so we need to disable that and skip some tests.
Annoying part is that it is also part of the `composer.json`, so we might need to remove that in a `before_script`:

```yml
before_script:
  ..
  - sed -r 's/^.*mongodb.*$//' composer.json > composer.json.copy; mv composer.json.copy composer.json
  ..
```

Maks3w:
* Cache composer dependencies.
* Speed up test execution (remove xdebug).
* ~~Add PHP 7 to test matrix.~~ "Tests are not executed"
* Add HHVM to test matrix.

Matrix execution time:
* Before: 19m 31s (last execution on master)
* Current:  6m 52s

Fix #76

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -20,7 +20,10 @@ matrix:
           env: COMPOSER_LOWEST=""
         - php: 5.6
           env: COMPOSER_LOWEST="--prefer-lowest"
+        - php: 7
     fast_finish: true
+    allow_failures:
+        - php: 7
 
 before_install:
     # Disable XDebug speed up test execution. (Ignore failures if platform does not had the extension installed)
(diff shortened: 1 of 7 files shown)


wjzijderveld:
Nice! 

I'm curious why PHP 7 is not cooperating there, but nice that we can test against HHVM with this.
If there is no chance on fixing PHP 7, please remove the commented lines.

:+1: from me

Maks3w:
I think keep the commented lines it's important in this case. 

If anyone create a PR adding that line you could merge the PR without review test build since Travis don't fail.

So comments lines are a reminder to you for to review the run it's correct.

PR #210, opened 2015-09-07
Remove composer.lock

Maks3w:
Composer.lock file is not good for libraries. Avoid to replicate real composer scenarios where the default behavior is to install the most recent version of each dependency.


wjzijderveld:
Can you try this with #209? To see what it means for the total time to run?
I added composer.lock because Travis failed pretty often with timeouts, and builds just took forever...

I'm :+1: on removing it again, if it doesn't give us more headaches :smile:

Maks3w:
@wjzijderveld Total time for #209 should be almost the same due the cache of dependencies.

Just it may take 4 or 5 seconds more for to check if the cached dependencie it's up to date.

Maks3w:
I've added a commit to #209 for to use `travis_retry` on Composer commands. `travis_retry` retry the command up to 3 times if fails.

rdohms:
I have been in many `composer.lock` arguments, so i'll weigh in on this.

The biggest motivation for not having the lock file is that travis will then "run free" and end up trying software combinations that you did not forsee, and may catch some unexpected or tested bugs.

However, this is very negative for contributors, because they may get test failures in their PR which they are not responsible for, and this leads to confusion on their end. IMHO i think contributors are really important and need to be respected.

I still think its valid to test "unusual" scenarios, but not in "random" form. So I recommend this blogpost http://blog.wyrihaximus.net/2015/06/test-lowest-current-and-highest-possible-on-travis/ as the best alternative IMO, gives you testing of unexpecteds, but also gives your contributors stability.

my 2 cents.

Xerkus:
@rdohms while it might be good for contributors, it will be also bad for consumers who are NOT affected by composer.lock

rdohms:
@Xerkus the lock file does not affect consumers at all. If you mean that consumers will not "get full coverage" then that is exactly what the blog post is for, it will test a sufficient amount of combinations, which i think is more the ideal.

Xerkus:
@rdohms approach testing combinations will require build to fail if current or highest fail, so it is kinda moot point anyway

Maks3w:
1. Test real dependency graph scenarios it's imperative. This is much important than any contributor situation.

2. Test lower and higher dependencies is alredy done in .travis,yml (at least for SF dependency)

3. PR unrelated errors will occur no matter if composer.lock exists. (Broken tests committed, changes in Travis infraestructure, Software versions, Network issues.) The repo maintainer should be responsible of fix those ones (where the codebase is involved)

rdohms:
@Xerkus not really, since it will be possible to identify that the failure happened due to an external condition, not due to the new code introduced in the PR, so contributor or repo owner have more information at hand, as opposed to having to figure it out by debugging.

rdohms:
@Maks3w why do you consider "loose dependency acquisition" to be "real" its random. It can also be replicated by higher, in a controller manner, with better and clearer intent being communicated.

Allowing one error to happen because "any error" can happen is not really justifiable now is it? PR unrelated issues should be mitigated for better testing context.

wjzijderveld:
@rdohms Thanks for your thoughts, I used the blogpost you refer to, to add coverage against Symfony 2.3. 

It probably is a good idea to change .travis.yml to do a full update with `--prefer-lowest` and add a case in the matrix to test against all latest versions as well (so a normal `composer update`).

With those changes, there aren't (IMO) big reasons left whether it should be in or out. So flip a coin? :smiling_imp: 
Trying to minimize PR unrelated errors is not a good reason for me, as this is meant to assure quality, not a happy contributor. We can help contributors in other ways, like using their solution but fix build errors ourselves, or by assisting them with the build errors.

Maks3w:
@wjzijderveld I have in my backlog to propose generalize the prefer-lowest but it's blocked until #209 is merged. (merge conflicts)

Maks3w:
rebased

othillo:
I'm in favor of removing. ATM most of us develop on PHP7 and commit their lock files. Then Travis CI fails because we test PHP 5.5 and 5.6.

PR #217, opened 2015-10-22
Fix saga state done setter

jsor:
Use `$state->setDone()` instead of ` $state->isDone()` to end the saga.


wjzijderveld:
Thanks!

PR #225, opened 2015-11-30
Fix date comparison

### LINKED ISSUE #203: DateTime does not use microseconds in comparison
See https://github.com/dbellettini/broadway/commit/705a7773f712f4ea1fdb7c6be79951c1547becdb

wjzijderveld:
Fixes #203 

[PHP doesn't include microseconds when comparing dates](http://lxr.php.net/xref/PHP_7_0/ext/date/php_date.c#2129), as it uses the seconds since epoch there.
To solve this in userland, we need to fall back to string comparison. To make that possible I had to force the internal timezone to UTC.
This might have side-effects if you used the fact that the DateTime object always returned in the TimeZone you created it in (when using `fromString`).


src/Broadway/Domain/DateTime.php
```diff
@@ -26,6 +26,7 @@ class DateTime
 
     private function __construct(BaseDateTime $dateTime)
     {
+        $dateTime->setTimezone(new DateTimeZone('UTC'));
```
asm89: We should at least mention this in the changelog.

We could also only force a copy to UTC when folks go comparing?

wjzijderveld:
I liked idea to only do the timezone conversion in a local copy, so it doesn't leak out but stays immutable.

asm89:
@wjzijderveld the cloning seems expensive and we're only looking at fixing the microsecond case here. I'm wondering if making this code a bit more complex makes sense for performance reasons?

```
if date1 > date2: # in this case performance stays the same
  return true
else if date1 < date2: # otherwise you get one extra check
  return false
else: # if all else fails you'll get the expensive check
  // do expensive compare
```

PR #228, opened 2015-12-14
Bundle is now Symfony 3 compatible

ricbra:
To use the bundle with Symfony 3 I had to remove some deprecated configuration options: <code>factory-class</code>, <code>factory-method</code>, <code>factory-service</code>.

These are deprecated since 2.6.

### DIFF
--- a/src/Broadway/Bundle/BroadwayBundle/Resources/config/saga/mongodb.xml
+++ b/src/Broadway/Bundle/BroadwayBundle/Resources/config/saga/mongodb.xml
@@ -9,11 +9,8 @@
             <argument type="service" id="broadway.saga.state.mongodb_collection" />
         </service>
 
-        <service id="broadway.saga.state.mongodb_database"
-                 class           = "Doctrine\MongoDB\Database"
-                 factory-service = "broadway.saga.state.mongodb_connection"
-                 factory-method  = "selectDatabase"
-                >
+        <service id="broadway.saga.state.mongodb_database" class="Doctrine\MongoDB\Database">
+            <factory service="broadway.saga.state.mongodb_connection" method="selectDatabase" />
             <argument>%broadway.saga.mongodb.database%</argument>
         </service>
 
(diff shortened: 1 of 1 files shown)


nilopc:
:clap:

rjkip:
Afaik this syntax is not supported before Symfony 2.6, so this change should be considered carefully. If Symfony <2.6 is supported by Broadway, this can be addressed through the extension ([example](https://github.com/doctrine/DoctrineBundle/blob/44b1ca7b82a56f1af2b1fbe7477b3b36c38ee2b7/DependencyInjection/DoctrineExtension.php#L135-L143)).

ricbra:
Don't know if Symfony < 2.6 is supported by broadway, but since Symfony 2.3 is maintained until 05/2016 we better fix this in the extension. 

@wjzijderveld what do you think?

othillo:
I'm in favor of dropping support for SF < 2.6 and merge this PR.
We do not test against SF < 2.6 anyway and I don't like the additional checks to support it as @rjkip points out.

@qandidate-labs/broadway do you agree?

Spy-Seth:
:+1:

wjzijderveld:
Thanks @ricbra 

When we split out the bundle (it will happen at some point! :yum:) we should make sure we specify the requirements properly.

mbadolato:
@wjzijderveld @othillo Does ```|~3.0``` needed to be added to the symfony components in ```composer.json``` as well? /me thinks the current restrictions to the ```~2.n``` branch would still keep broadway from being added to a Symfony project

ricbra:
Current requirements don't specify any symfony component:

```json
{
    "require": {
        "beberlei/assert": "~2.0",
        "broadway/uuid-generator": "~0.1.0",
        "php": ">=5.5.9",
        "ramsey/uuid": "~2.4"
    },
}
```

Only the dev requirements specify some symfony components, so that's not a problem. But it would be great if the bundle could have its own dedicated package (with composer.json) as @wjzijderveld mentioned.

mbadolato:
the require-dev isn't as much of a problem for production but it would be a problem if we have dev requirements installed on local boxen when developing the app

wjzijderveld:
require-dev isn't recursive. So if your application includes Broadway, even in dev, it wouldn't install symfony neither will it check the constraint. Only when you actually clone Broadway and run composer install in the Broadway directory it will check require-dev

mbadolato:
Ah! Didn't know that. Thx

PR #230, opened 2016-01-03
Allow Serializer to run recursively

remi-san:
No Breaking Changes

The serializer can serialize on multiple levels if the serialized object contains another serializable object.

Tests have been added to test the added feature.

### DIFF
--- a/src/Broadway/Serializer/SimpleInterfaceSerializer.php
+++ b/src/Broadway/Serializer/SimpleInterfaceSerializer.php
@@ -32,8 +32,26 @@ class SimpleInterfaceSerializer implements SerializerInterface
 
         return array(
             'class'   => get_class($object),
-            'payload' => $object->serialize()
+            'payload' => $this->serializePayload($object->serialize())
         );
+    }
+
+    /**
+     * Recursively serialize the object
+     *
+     * @param  array $serialized
+     * @return array
+     */
+    private function serializePayload(array $serialized)
+    {
+        $payload = array();
+
+        foreach ($serialized as $name => $serializedItem) {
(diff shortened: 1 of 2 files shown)


asm89:
:-1: this makes the serializer too complex imo. The `serialize()` method should return an array of primitives.

What is your use case for this?

In general I'd recommend folks with more complicated serialization needs to hook up something like jms serializer.

remi-san:
I need it to serialize the Metadata object for instance.

I don't really want to extend the existing class, but I need it to contain complex objects. But the serialize method only returns the array of parameters, which is ok when you have primitives, but won't serialize my inner objects.

I don't think it adds a lot of complexity, and if you don't want the complexity, you can return arrays of primitives, it won't change anything. As I said, there's no breaking changes.

Furthermore, I don't think that making a class know how to serialize another class is a good idea.

And JMS is a very heavy solution for a simple (but very ordinary) case like this one.

mbadolato:
@remi-san Would something like Mattias Noback's [Broadway Serialization helper library](https://github.com/matthiasnoback/broadway-serialization) be helpful for your situation?

Each value object would need to implement the library's interface and provide the rules, but it's pretty simple to use and allows for recursive serializing/deserializing. I haven't used it beyond testing some things, but seemed to work pretty well when I did that.

remi-san:
I think that should do the trick, yes.

But I'd still have to create a new class of Metadata, which was what I wanted to avoid.

Nonetheless, thank you for making me discover that lib. I might use it later.

It's just that I had this developed for one of my projects and wanted to share it. But I understand the reasons why you wouldn't want that.

wjzijderveld:
Closing for reasons mentioned above

PR #236, opened 2016-02-01
Replace doctrine mongo

wjzijderveld:
- [x] Refactor StateRepository to use `mongodb` extension
- [ ] Refactor BroadwayBundle services

I still want to quickly look into the possibility to support both extensions (`mongo` and `mongodb`, that would make upgrading a lot easier. Although you can install both extensions next to each other, so it's not really required.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -33,7 +33,7 @@ addons:
 before_install:
     # Disable XDebug speed up test execution. (Ignore failures if platform does not had the extension installed)
     - phpenv config-rm xdebug.ini || return 0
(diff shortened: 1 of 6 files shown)


wjzijderveld:
Hmm, travis doesn't have mongodb.so, but [this page](https://docs.travis-ci.com/user/languages/php/#PHP-extensions) says it should have it...

othillo:
see https://github.com/broadway/broadway/pull/279/commits/90e02e4f0329ce4d5bd21615347648a83b16fc90

wjzijderveld:
We'll go with the polyfill now

PR #238, opened 2016-02-26
[RFC] create alias when creating Elasticsearch indices

othillo:


asm89:
Should we make creating an alias yes/no configurable for BC purposes?

othillo:
moved to https://github.com/broadway/read-model-elasticsearch/issues/2

Issue #244, opened 2016-03-09
Reduce code duplication SimpleEventBus


DawidMazurek:
Make use of _finally_ to prevent code duplication in _try_ and _catch_ blocks in SimpleEventBus

PR #245, opened 2016-03-09
Reducing code duplication by moving duplicated logic from try and cat…

DawidMazurek:
…ch to single finally block

#244

### DIFF
--- a/src/Broadway/EventHandling/SimpleEventBus.php
+++ b/src/Broadway/EventHandling/SimpleEventBus.php
@@ -49,11 +49,11 @@ class SimpleEventBus implements EventBusInterface
                         $eventListener->handle($domainMessage);
                     }
                 }
-
-                $this->isPublishing = false;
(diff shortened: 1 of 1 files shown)


src/Broadway/EventHandling/SimpleEventBus.php
```diff
@@ -49,12 +49,12 @@ public function publish(DomainEventStreamInterface $domainMessages)
                         $eventListener->handle($domainMessage);
                     }
                 }
-
-                $this->isPublishing = false;
             } catch (Exception $e) {
-                $this->isPublishing = false;
                 throw $e;
             }
+            finally {
```
asm89: nit: `} finally {`

asm89:
lgtm :+1:

GromNaN:
The catch part is not necessary.

wjzijderveld:
Thanks @DawidMazurek

PR #248, opened 2016-03-17
Serializers are now configurable via Bundle

ricbra:
I've implemented a POC for overriding or configuring the serializers. Why? There are some lazy people out there who don't want to implement `Broadway\Serializer\SerializableInterface` for every event and throw some magic add it :wink: .

I'd love to get some feedback on the current implementation.


othillo:
thank you, @ricbra !

PR #252, opened 2016-03-29
Update EventStreamNotFoundException  add tablename

Miliooo:
I'm using different event tables for different parts of my site. So having a bit more info in the exception is quite helpful.


wjzijderveld:
Thanks @Miliooo

PR #258, opened 2016-04-20
Specifying when events happen for ReadModel Scenario

### LINKED ISSUE #257: Cannot specify DateTime for ReadModel Scenario's given
The [ReadModel Scenario](https://github.com/qandidate-labs/broadway/blob/c2939cdd68edf1f44cdaef3ce0288fd1250e18d1/src/Broadway/ReadModel/Testing/Scenario.php#L82) allows specifying when an event happened for the 'when' helper

However, this is not supported for the [given](https://github.com/qandidate-labs/broadway/blob/c2939cdd68edf1f44cdaef3ce0288fd1250e18d1/src/Broadway/ReadModel/Testing/Scenario.php#L68) method.

The issue we are facing is that we are using the DateTime from the DomainMessage in one of our projectors but we have no control over [the value being generated](https://github.com/qandidate-labs/broadway/blob/c2939cdd68edf1f44cdaef3ce0288fd1250e18d1/src/Broadway/ReadModel/Testing/Scenario.php#L106) by the scenario

fritsjanb:
Fixes #257

### DIFF
--- a/src/Broadway/ReadModel/Testing/Scenario.php
+++ b/src/Broadway/ReadModel/Testing/Scenario.php
@@ -35,19 +35,21 @@ class Scenario
     private $repository;
     private $playhead;
     private $aggregateId;
-    private $dateTime;
+    private $dateTimeGenerator;
 
     public function __construct(
         PHPUnit_Framework_TestCase $testCase,
         RepositoryInterface $repository,
         ProjectorInterface $projector
     ) {
-        $this->testCase    = $testCase;
-        $this->repository  = $repository;
-        $this->projector   = $projector;
-        $this->playhead    = -1;
-        $this->aggregateId = 1;
-        $this->dateTime    = DateTime::now();
+        $this->testCase          = $testCase;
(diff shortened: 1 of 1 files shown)


src/Broadway/ReadModel/Testing/Scenario.php
```diff
@@ -60,6 +62,11 @@ public function withAggregateId($aggregateId)
         return $this;
     }
 
+    public function usingDateTime(DateTime $dateTime)
```
asm89: `withDateTime()` and return `$this`?

asm89:
I'm conflicted about this solution, because:

- the datetime object will be the exact same object for each event (opposed to a new object each time)
- the timing for each events in the tests will also be the same

I understand that under most conditions the latter is already the case, but it's something we should solve instead of mask further.

asm89:
lgtm

mente:
Why not use the same approach as it was for `when()`?

PR #260, opened 2016-05-02
Add Metadata to DomainMessage for Unit Tests

Konafets:
When creating an event, Broadway gives you the possibility to enrich the DomainMessage with metadata. This is done via Decorators and Enrichers. However, in Unit Testing there is no way to mock metadata because no Decorators are called. To mock this behavior for testing, just add a new parameter to the when()-method which gives the possibility to pass the Metadata further down to the DomainMessage.

Background:
In some cases it is necessary to access the stored metadata when applying the Events onto your ReadModel. If the stored data is not available in testing context, the tests fails with an "Index not found"
exception.

### DIFF
--- a/src/Broadway/ReadModel/Testing/Scenario.php
+++ b/src/Broadway/ReadModel/Testing/Scenario.php
@@ -75,14 +75,32 @@ class Scenario
     }
 
     /**
-     * @param array $events
+     * Add events in the form of
+     *
+     * - no events: given([])
+     * - one event: given(new CustomerWasCreatedEvent(...)) or given([new CustomerWasCreatedEvent(...)])
+     * - event with metadata: given(['event' => new CustomerWasCreatedEvent(...), 'metadata' => ['foo' => 'bar']])
      *
      * @return Scenario
      */
-    public function given(array $events = array())
+    public function given()
     {
-        foreach ($events as $given) {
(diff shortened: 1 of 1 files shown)


wjzijderveld:
I understand your issue, but I'm not convinced about this solution. Seems a bit to hacky :-)

Maybe it would be better if we add the possibility to pass DomainMessages instead of events? 
I wouldn't want to give up the ease of use we have right now, so I would vote for a new method to handle those cases.

```php
public function givenDomainMessages(array $domainMessages) // or DomainEventStreamInterface $domainMessages
``` 

An alternative would be to keep it in 1 method, but only wrap the event if it's not already a DomainMessage, but imo that is still a bit hacky.

```php
public function given(array $messages)
{
  $messages = array_map(function ($message) {
    if (! $message instanceof DomainMessage) {
      $message = $this->createDomainMessageForEvent($message);
    }

    return $message;
  }, $messages);
  
  // ...
}
```

@qandidate-labs/broadway Any thoughts?

othillo:
I prefer passing DomainMessages. For backward compatibility I like the givenDomainMessages solution.

othillo:
@Konafets does this solve your problem? https://github.com/broadway/broadway/pull/300

Feel free to reopen.

PR #264, opened 2016-05-16
Adds ability to customize DomainMessage getType method

aleixverges:
- Creates an EventStream decorator to allow DomainMessage type customization
- Updates DomainMessage to allow type update

### DIFF
--- a/src/Broadway/Domain/DomainMessage.php
+++ b/src/Broadway/Domain/DomainMessage.php
@@ -87,7 +87,7 @@ final class DomainMessage
     }
 
     /**
-     * {@inheritDoc}
+     * @return mixed | EventInterface
      */
     public function getPayload()
     {
@@ -147,8 +147,9 @@ final class DomainMessage
      */
     public function andType($type)
     {
-        $this->type = $type;
+        $message = clone $this;
+        $message->type = $type;
 
-        return $this;
+        return $message;
     }
 }
--- a/src/Broadway/Domain/DomainMessage.php
+++ b/src/Broadway/Domain/DomainMessage.php
@@ -42,6 +42,11 @@ final class DomainMessage
     private $recordedOn;
 
     /**
+     * @var string
+     */
+    private $type;
+
+    /**
      * @param string   $id
      * @param int      $playhead
      * @param Metadata $metadata
@@ -102,7 +107,11 @@ final class DomainMessage
      */
     public function getType()
     {
-        return strtr(get_class($this->payload), '\\', '.');
+        if (!$this->type) {
+            return strtr(get_class($this->payload), '\\', '.');
+        }
+
+        return $this->type;
     }
 
     /**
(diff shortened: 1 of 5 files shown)


src/Broadway/EventSourcing/DomainMessageTypeEnrichment/DomainMessageTypeEventStreamDecorator.php
```diff
@@ -0,0 +1,29 @@
...
+    /**
+     * {@inheritDoc}
+     */
+    public function decorateForWrite($aggregateType, $aggregateIdentifier, DomainEventStreamInterface $eventStream)
+    {
+        $messages = array();
+
+        foreach ($eventStream as $message) {
+            $payload = $message->getPayload();
+            if (method_exists($payload, 'getType')) {
```
unkind: What if event accidentally contains this getter?

src/Broadway/Domain/DomainMessage.php
```diff
@@ -131,4 +140,15 @@ public function andMetadata(Metadata $metadata)
...
         return new DomainMessage($this->id, $this->playhead, $newMetadata, $this->payload, $this->recordedOn);
     }
+
+    /**
+     * @param string $type
+     * @return $this
+     */
+    public function andType($type)
+    {
+        $this->type = $type;
```
unkind: This makes `DomainMessage` mutable.

src/Broadway/EventSourcing/DomainMessageTypeEnrichment/DomainMessageTypeEventStreamDecorator.php
aleixverges: That's more less the plan. A new EventInterface could exist with the getType or messageName method defined. The decorator would use that to create new DomainMessage with that new message key.

src/Broadway/Domain/DomainMessage.php
aleixverges: A new optional parameter could be added to DomainMessage class called "type" for example. That would make the creation of a new DomainMessage, like with the andMetadata decorator, possible

aleixverges:
Thanks for the feedback!
      I can prepare an updated pull request to show the changes in the implementation fixing the issues you saw.

Aleix

src/Broadway/Domain/DomainMessage.php
unkind: You don't have to use constructor:
```
public function andType($type)
{
    $message = clone $this;
    $message->type = $type;

    return $message;
}
```

aleixverges:
Hi,
    I've updated the pull request with some changes according your comments.
Thanks!
Aleix

src/Broadway/EventSourcing/DomainMessageTypeEnrichment/DomainMessageTypeEventStreamDecorator.php
```diff
@@ -0,0 +1,31 @@
...
+     * {@inheritDoc}
+     */
+    public function decorateForWrite($aggregateType, $aggregateIdentifier, DomainEventStreamInterface $eventStream)
+    {
+        $messages = array();
+
+        /** @var DomainMessage $message */
+        foreach ($eventStream as $message) {
+            $payload = $message->getPayload();
+            if (method_exists($payload, 'getType')) {
```
othillo: if ($payload implements EventInterface) ?

PR #265, opened 2016-05-24
Add ReflectionAggregateFactory

unkind:
Hi,

In this PR I've added factory which doesn't call constructor. There is already similar factory `NamedConstructorAggregateFactory`, but I don't like it for 2 reasons:
* it forces to have technical named constructor in my shiny model;
* that technical constructor calls regular constructor; but in my opinion it's better to call regular one only when entity's lifecycle starts, but reconstitution is not start of lifecycle.

### DIFF
--- a/test/Broadway/EventSourcing/AggregateFactory/ReflectionAggregateFactoryTest.php
+++ b/test/Broadway/EventSourcing/AggregateFactory/ReflectionAggregateFactoryTest.php
@@ -1,0 +1,85 @@
+<?php
+
+namespace Broadway\EventSourcing\AggregateFactory;
+
+use Broadway\Domain\DomainEventStream;
+use Broadway\EventSourcing\EventSourcedAggregateRoot;
+use Broadway\TestCase;
+
+/**
+ *
+ */
+final class ReflectionAggregateFactoryTest extends TestCase
+{
+    /**
+     * @var ReflectionAggregateFactory
+     */
+    private $factory;
+
+    public function setUp()
+    {
+        $this->factory = new ReflectionAggregateFactory();
+    }
+
+    /**
+     * @test
+     */
+    public function it_creates_instance_of_aggregate_with_private_constructor()
+    {
+        $aggregate = $this->factory->create(
+            TestAggregateWithPrivateConstructor::class,
+            new DomainEventStream([])
+        );
+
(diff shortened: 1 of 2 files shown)


test/Broadway/EventSourcing/AggregateFactory/ReflectionAggregateFactoryTest.php
```diff
@@ -0,0 +1,63 @@
...
+     * @expectedException \LogicException
+     * @expectedExceptionMessage Impossible to initialize "stdClass"
+     */
+    public function it_does_not_handle_weird_classes()
+    {
+        $this->factory->create(\stdClass::class, new DomainEventStream([]));
+    }
+}
+
+// TODO: separate folder with test aggregates?
```
wjzijderveld: I think it makes the test more clear to keep them in the same file for now.
If we would get more complex tests, we might want to move them somewhere else.

For now I would say, remove this TODO and I'll merge this :)

test/Broadway/EventSourcing/AggregateFactory/ReflectionAggregateFactoryTest.php
```diff
@@ -0,0 +1,63 @@
...
+
+final class MyTestAggregateRootForReflectionFactory extends EventSourcedAggregateRoot
+{
+    private function __construct()
```
asm89: Does it work if you don't define a constructor at all? Perhaps add a separate test for it.

test/Broadway/EventSourcing/AggregateFactory/ReflectionAggregateFactoryTest.php
unkind: > I think it makes the test more clear to keep them in the same file for now.
I worry about name conflicts with other stubs.

Anonymous classes would be perfect replacement.

wjzijderveld: Yeah, that is a risk indeed, Although I think it won't be that big of a deal.
Most of these dummies are quit specific for a testcase so can have a very specific name.

Let's at least not try to solve that in this PR. Feel free to open a new PR if you want to :-)

wjzijderveld:
Thanks @unkind !

PR #267, opened 2016-06-15
Duplicate playhead

### LINKED ISSUE #266: No way to detect race conditions on even stream
Hi,

We are working on a system that requires an exclusive lock, before calling an external api. In order to do that we append an event to the store to ensure the process is not performed twice.

However in case of a race condition, the current EventStore throws a generic exception, which is hard to handle, because it can pretty much mean anything at that point, and we have to fall back to string matching.

This is what I propose:
- Add EventStoreException to @throws in DocBlock of EventStore::append (This should already be the case imo)
- Introduce specific exception for race conditions in event store that extends EventStoreException.
- The only regression would be if people catch `DBALEventStoreException` instead of `EventStoreException`.

The question is how should the exception be named, this is what I could think of:
- `DuplicatePlayheadException`
- `EventRaceConditionException`
- `OutdatedAggregateException`

I'm aware of the sensible interface discussion, but I think it should be out of scope for now to prevent an inconsistent base.

If you guys like this I will throw in a PR. Please let me know what class name you prefer.

reenl:
Allow the domain to detect race conditions on the eventstream, without depending on a driver implementation.

BC break (Documented in `CHANGELOG.md`):
If an application would catch `DBALEventStoreException` the `DuplicatePlayheadException` will be uncaught.

This can be fixed by introducing more classes and interfaces, however since we are on a 0.x release I didn't create that overhead. 

If this gets merged, people should update their code to only catch `EventStoreException`s and not the specific `DBALEventStoreException`, catching it would break the SOLID principle anyway, because you depend on implementation and not on the abstraction.

Resolves #266

### DIFF
--- a/src/Broadway/EventStore/EventStoreInterface.php
+++ b/src/Broadway/EventStore/EventStoreInterface.php
@@ -12,6 +12,7 @@
 namespace Broadway\EventStore;
 
 use Broadway\Domain\DomainEventStreamInterface;
+use Broadway\EventStore\Exception\DuplicatePlayheadException;
 
 /**
  * Loads and stores events.
@@ -28,6 +29,8 @@ interface EventStoreInterface
     /**
      * @param mixed                      $id
      * @param DomainEventStreamInterface $eventStream
+     *
+     * @throws DuplicatePlayheadException
      */
     public function append($id, DomainEventStreamInterface $eventStream);
 }
--- a/src/Broadway/EventStore/Exception/DuplicatePlayheadException.php
+++ b/src/Broadway/EventStore/Exception/DuplicatePlayheadException.php
@@ -1,0 +1,14 @@
+<?php
+
+namespace Broadway\EventStore\Exception;
+
+use Broadway\EventStore\EventStoreException;
+use Exception;
+
+class DuplicatePlayheadException extends EventStoreException
+{
+    public static function create(Exception $exception)
+    {
+        return new self(null, 0, $exception);
+    }
+}
--- a/src/Broadway/EventStore/Exception/DuplicatePlayheadException.php
+++ b/src/Broadway/EventStore/Exception/DuplicatePlayheadException.php
@@ -2,13 +2,33 @@
 
 namespace Broadway\EventStore\Exception;
 
+use Broadway\Domain\DomainEventStreamInterface;
 use Broadway\EventStore\EventStoreException;
 use Exception;
 
 class DuplicatePlayheadException extends EventStoreException
 {
-    public static function create(Exception $exception)
+    /**
+     * @var DomainEventStreamInterface
+     */
+    private $eventStream;
+
+    /**
+     * @param DomainEventStreamInterface $eventStream
+     * @param Exception                  $previous
+     */
+    public function __construct(DomainEventStreamInterface $eventStream, $previous = null)
     {
-        return new self(null, 0, $exception);
+        parent::__construct(null, 0, $previous);
+
+        $this->eventStream = $eventStream;
+    }
+
+    /**
+     * @return DomainEventStreamInterface
+     */
+    public function getEventStream()
+    {
+        return $this->eventStream;
     }
 }
--- a/src/Broadway/EventStore/InMemoryEventStore.php
+++ b/src/Broadway/EventStore/InMemoryEventStore.php
@@ -13,6 +13,7 @@ namespace Broadway\EventStore;
 
 use Broadway\Domain\DomainEventStream;
 use Broadway\Domain\DomainEventStreamInterface;
+use Broadway\EventStore\Exception\DuplicatePlayheadException;
 use Broadway\EventStore\Management\Criteria;
 use Broadway\EventStore\Management\EventStoreManagementInterface;
 
@@ -61,7 +62,7 @@ class InMemoryEventStore implements EventStoreInterface, EventStoreManagementInt
     private function assertPlayhead($events, $playhead)
     {
         if (isset($events[$playhead])) {
-            throw new InMemoryEventStoreException(
+            throw new DuplicatePlayheadException(
                 sprintf("An event with playhead '%d' is already committed.", $playhead)
             );
         }
--- a/src/Broadway/EventStore/InMemoryEventStore.php
+++ b/src/Broadway/EventStore/InMemoryEventStore.php
@@ -13,6 +13,7 @@ namespace Broadway\EventStore;
 
 use Broadway\Domain\DomainEventStream;
 use Broadway\Domain\DomainEventStreamInterface;
+use Broadway\Domain\DomainMessage;
 use Broadway\EventStore\Exception\DuplicatePlayheadException;
 use Broadway\EventStore\Management\Criteria;
 use Broadway\EventStore\Management\EventStoreManagementInterface;
@@ -51,20 +52,29 @@ class InMemoryEventStore implements EventStoreInterface, EventStoreManagementInt
             $this->events[$id] = [];
         }
 
+        $this->assertStream($this->events[$id], $eventStream);
+
+        /** @var DomainMessage $event */
         foreach ($eventStream as $event) {
             $playhead = $event->getPlayhead();
-            $this->assertPlayhead($this->events[$id], $playhead);
 
             $this->events[$id][$playhead] = $event;
         }
     }
 
-    private function assertPlayhead($events, $playhead)
+    /**
+     * @param DomainMessage[]            $events
+     * @param DomainEventStreamInterface $eventsToAppend
+     */
+    private function assertStream($events, $eventsToAppend)
     {
-        if (isset($events[$playhead])) {
-            throw new DuplicatePlayheadException(
-                sprintf("An event with playhead '%d' is already committed.", $playhead)
-            );
+        /** @var DomainMessage $event */
+        foreach ($eventsToAppend as $event) {
+            $playhead = $event->getPlayhead();
+
+            if (isset($events[$playhead])) {
+                throw new DuplicatePlayheadException($eventsToAppend);
+            }
         }
     }
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,8 @@
 - We moved the Symfony bundle to a [separate repository](https://github.com/broadway/broadway-bundle)
 - The DBALEventStore requires a BinaryUuidConverter
 - We moved the Saga component to a [separate repository](https://github.com/broadway/broadway-saga)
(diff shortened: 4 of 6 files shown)


src/Broadway/EventStore/EventStoreInterface.php
```diff
@@ -28,6 +28,8 @@ public function load($id);
     /**
      * @param mixed                      $id
      * @param DomainEventStreamInterface $eventStream
+     *
+     * @throws EventStoreException
```
unkind: I think it's better to specify concrete exceptions, i.e. `@throws DuplicatePlayheadException`.

src/Broadway/EventStore/Exception/DuplicatePlayheadException.php
```diff
@@ -0,0 +1,14 @@
+<?php
+
+namespace Broadway\EventStore\Exception;
+
+use Broadway\EventStore\EventStoreException;
+use Exception;
+
+class DuplicatePlayheadException extends EventStoreException
+{
+    public static function create(Exception $exception)
```
unkind: `withPrevious`?

src/Broadway/EventStore/EventStoreInterface.php
reenl: This was added because the exception was undocumented, its more of a bugfix.

I can add both, just to clearify?

src/Broadway/EventStore/Exception/DuplicatePlayheadException.php
reenl: Yeah I had a really hard time here, the codebase isn't really consistent. However this was based on the DBALEventStoreException api.

I'll be happy to change it, when we agree upon format.

Thanks for the review @unkind

src/Broadway/EventStore/EventStoreInterface.php
unkind: In my opinion, `DuplicatePlayheadException` should be documented explicitly in this method.

On other hand, some driver-specific exceptions like "Database is down" don't make much sense from `EventStore` client perspective. Someone who uses this interface shouldn't be aware of such implementation details.

Generic `EventStoreException` doesn't tell what happened exactly, so I don't think someone would catch it.

By analogy with Java checked/unchecked exceptions, I'd say `DuplicatePlayheadException` is checked one and "Database is down" is unchecked one. So I'd leave `@throws DuplicatePlayheadException` alone.

reenl: Makes sense to me. Since changes are so minor I'll await feedback from qandidate before fixing it.

Also thanks for referencing to "(un)checked" exceptions, didn't know this concept had a name :)

src/Broadway/EventStore/Exception/DuplicatePlayheadException.php
wjzijderveld: To be honest, I'm not even sure if you would want to assume this exception is always based on another exception. For event stores where the duplicate isn't detected based on an exception, you would need to create an exception to create an exception...

I would prefer to have a method where you pass the id and playhead that are causing the duplicate (and pass an optional previous exception), but that will be tricky as you can't really get that info from the DBALException.... But maybe we can at least report for which identifier the duplicate occurred?

src/Broadway/EventStore/EventStoreInterface.php
wjzijderveld: I'm fine either way, as PHP doesn't actually have the concept of (un)checked exceptions, it might be nice to still document that we can throw both exceptions. But I do agree that it's an exception you probably can't recover from.

wjzijderveld:
Thanks for the work so far @reenl!

src/Broadway/EventStore/EventStoreInterface.php
unkind: What's the benefit of specifying `EventStoreException` in the interface? In my opinion it makes API harder to understand.

src/Broadway/EventStore/Exception/DuplicatePlayheadException.php
reenl: I like the idea, and it would actually be pretty easy to implement. However, if we are considering this, shouldn't we just provide the uncommitted events?

wjzijderveld: Good idea, it is a transaction, so providing the complete transaction (uncommtted events) makes a lot of sense!

reenl:
@wjzijderveld You're welcome, just did some changes based on your feedback, let me know what you think.

src/Broadway/EventStore/InMemoryEventStore.php
```diff
@@ -46,25 +48,35 @@ public function append($id, DomainEventStreamInterface $eventStream)
...
+        if (isset($this->events[$id])) {
+            $aggregateEvents = $this->events[$id];
         }
 
         foreach ($eventStream as $event) {
-            $playhead = $event->getPlayhead();
-            $this->assertPlayhead($this->events[$id], $playhead);
-
-            $this->events[$id][$playhead] = $event;
+            $aggregateEvents = $this->insertEvent($aggregateEvents, $event, $eventStream);
```
wjzijderveld: I don't think the code get's any easier by extracting it to a method like this, I event think it makes it a lot more harder to understand then needed :)

src/Broadway/EventStore/InMemoryEventStore.php
```diff
@@ -46,25 +48,35 @@ public function append($id, DomainEventStreamInterface $eventStream)
...
         }
+
+        $this->events[$id] = $aggregateEvents;
```
wjzijderveld: Needed to look a few times before realising you are solving a bug with this :smile: 

I did note however that there is still a minor bug present, but I'll create a separate PR for that.

reenl: Sorry I'm sure I wrote this down, think commits got messed up. Can you reference / explain the bug?

src/Broadway/EventStore/InMemoryEventStore.php
reenl: Makes sense that leaves 2 options:
- Assert the complete stream instead of the event.
- Keep it the way it was, however the stream would be an extra argument to assertPlayhead.

While I'm at it I can also move the "transactional"-ish behaviour in another commit if you like.

src/Broadway/EventStore/InMemoryEventStore.php
wjzijderveld: Sure, it is a bug that is already present.

When providing an empty event stream, it will initialise `$this->events` with an empty array. This will cause the isset in `load` to succeed, which on its turn will return an empty event stream. So unless you guard for that in your aggregate, you will end up with an empty (and thus invalid)  aggregate. 
I think `load` should also check for array count instead of just isset.

reenl:
Allright so about the InMemoryEventStore, It will no longer apply any event if one of the playheads is known. (Transactional-ish)

Maybe we don't need to verify every single playhead, only the first, because they should be incremental. However changing this behaviour is out of scope for this PR.

janwalther:
What is the status of this feature? 
I would like to implement a conflict management as described in http://danielwhittaker.me/2014/09/29/handling-concurrency-issues-cqrs-event-sourced-system/ to save conflict exceptions of not really conflicting events. But I cannot find a way with the current `DBALEventStore`

reenl:
Hi @janwalther, No idea, I'll rebase and push to at least remove the conflicts.

reenl:
If there is anything else I can do please let me know, from my perspective this is mergable.

janwalther:
Looking forward for the merge. I already implemented a `ConflictResolvingEventStore`. Will create the full PR when this PR got merged.

In short the `ConflictResolvingEventStore` catches the `DuplicatePlayheadException` introduced in this PR, loads intersecting events (already committed events whose playhead is equal or greater than those of uncommitted events) and checks if the intersecting events really conflict with each other.

Preview:
```php
use Broadway\Domain\DomainEventStream;
use Broadway\Domain\DomainEventStreamInterface;
use Broadway\Domain\DomainMessage;
use Broadway\EventStore\EventStoreInterface;

class ConflictResolvingEventStore implements EventStoreInterface
{
    /** @var EventStoreInterface */
    private $eventStore;

    /** @var ConflictResolver */
    private $conflictResolver;

    /**
     * @inheritDoc
     */
    public function append($id, DomainEventStreamInterface $uncommittedEvents)
    {
        try {
            $this->eventStore->append($id, $uncommittedEvents);
        } catch (DuplicatePlayheadException $e) {
            $committedEvents = $this->eventStore->load($id);
            $conflictingEvents = $this->getConflictingEvents($uncommittedEvents, $committedEvents);

            $conflictResolvedEvents = [];
            $playhead = $this->getCurrentPlayhead($committedEvents);

            /** @var DomainMessage $uncommittedEvent */
            foreach ($uncommittedEvents as $uncommittedEvent) {
                foreach ($conflictingEvents as $conflictingEvent) {
                    if ($this->conflictResolver->conflictsWith($conflictingEvent, $uncommittedEvent)) {
                        throw $e;
                    }
                }

                $playhead++;

                $conflictResolvedEvents[] = new DomainMessage(
                    $id,
                    $playhead,
                    $uncommittedEvent->getMetadata(),
                    $uncommittedEvent->getPayload(),
                    $uncommittedEvent->getRecordedOn());
            }

            $this->append($id, new DomainEventStream($conflictResolvedEvents));
        }
    }

    /**
     * @inheritDoc
     */
    public function load($id)
    {
        $this->eventStore->load($id);
    }

    /**
     * @return int
     */
    private function getCurrentPlayhead(DomainEventStreamInterface $committedEvents)
    {
        $playhead = 0;

        foreach ($committedEvents as $committedEvent) {
            /** @var DomainMessage $committedEvent */
            $playhead = $committedEvent->getPlayhead();
        }

        return $playhead;
    }

    /**
     * @return DomainMessage[]
     */
    private function getConflictingEvents(DomainEventStreamInterface $uncommittedEvents, DomainEventStreamInterface $committedEvents) {
        $conflictingEvents = [];

        /** @var DomainMessage $committedEvent */
        foreach ($committedEvents as $committedEvent) {
            /** @var DomainMessage $uncommittedEvent */
            foreach ($uncommittedEvents as $uncommittedEvent) {
                if ($committedEvent->getPlayhead() >= $uncommittedEvent->getPlayhead()) {
                    $conflictingEvents[] = $committedEvent;

                    break;
                }
            }
        }

        return $conflictingEvents;
    }
}
`

wjzijderveld:
Thanks @reenl !

PR #274, opened 2016-10-31
Custom services for store and repository

robinvdvleuten:
Hi there, 

To use anything other than DBAL or ElasticSearch, the bundle must provide some way to have custom services configured. This is a simple change in the configuration / extension. Hope you guys like the idea :)

Cheers,
Robin


othillo:
Hi @robinvdvleuten, we had a discussion about this and we decided to adopt the proposal by @MattKetmo in https://github.com/broadway/broadway/issues/227 (which still needs to be built).

othillo:
@robinvdvleuten thanks for your POC! We haven't decided which approach to take. We will continue to discuss it here: https://github.com/broadway/broadway-bundle/issues/9

PR #279, opened 2016-12-07
remove dependency on ramsey/uuid

othillo:
fixes https://github.com/broadway/broadway/issues/255
fixes https://github.com/broadway/broadway/issues/263

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -5,9 +5,8 @@
     "license": "MIT",
     "require": {
         "beberlei/assert": "~2.0",
-        "broadway/uuid-generator": "~0.1.0",
-        "php": ">=5.5.9",
-        "ramsey/uuid": "~2.4"
+        "broadway/uuid-generator": "~0.3.0",
+        "php": ">=5.5.9"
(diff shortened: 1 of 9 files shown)


composer.json
```diff
@@ -43,7 +42,8 @@
         "monolog/monolog": "~1.8",
         "symfony/console": "~2.4",
         "symfony/proxy-manager-bridge": "~2.4",
-        "phpunit/phpunit": "^4.8"
+        "phpunit/phpunit": "^4.8",
+        "ramsey/uuid": "^3.3"
```
wjzijderveld: Why do we still need this in dev? It seems you removed all references already?

wjzijderveld:
Can't we get rid of the `--ignore-platform-reqs` in `.travis.yml`?

othillo:
I've removed the removal of composer.lock from this PR.

othillo:
ramsey/uuid is still a dev dependency as the tests use the concrete binary uuid converter

PR #287, opened 2016-12-20
[RFC] SerializableReadModelInterface

othillo:
Most of the time you want your read model to be serializable. In fact, the ElasticSearchRepository requires the read model to be serializable when using the default serializer. Also the https://github.com/broadway/broadway/blob/master/src/Broadway/ReadModel/Testing/ReadModelTestCase.php expects the read model to be serializable.

This interface helps making both behaviours explicit.

see https://github.com/broadway/broadway/issues/172


wjzijderveld:
Nice one, thanks!

PR #289, opened 2017-01-23
Concurrency conflict resolver

janwalther:
This additional event store decorates any other event store so you can define event types which are independent. This means that the order of those events does not matter. In case of a DuplicatePlayheadException the playhead of idependent events gets recalculated and thus the conflict can get solved.

For example for a User aggregate you could have EmailAddressChanged and UsernameChangedEvent. While the order of two UsernameChangedEvents does matter (because the last stored event defines the current username), it does not matter if a UsernameChangedEvent or a EmailAddressChanged event gets stored first.

Idea is based on [Daniel Whittaker's article on concurrency handling](http://danielwhittaker.me/2014/09/29/handling-concurrency-issues-cqrs-event-sourced-system/)


janwalther:
Currently the original test `EventStoreTest::it_throws_an_exception_when_appending_a_duplicate_playhead` fails. I think the reason is the wrong definition of the ConcurrencyConflictResolver mock but currently I do not know how to fix it. Any help appreciated.

othillo:
thank you, @janwalther !

PR #291, opened 2017-01-30
Update ElasticSearchRepository.php

raistlin:
Please see https://www.elastic.co/guide/en/elasticsearch/client/php-api/current/php_json_objects.html


othillo:
thank you, @raistlin !

PR #293, opened 2017-02-10
Idempotent commands

janwalther:
This PR enables Broadway users to ensure that each command is only executed once. Therefore it adds a command id to the domain messages' meta data with an event stream decorator. This command id gets transferred together with the request in a real-world app. If the same command gets executed a second time a DuplicateCommandException will be thrown.

### DIFF
--- a/src/Broadway/EventSourcing/IdempotentCommands/IdempotentCommandStreamDecorator.php
+++ b/src/Broadway/EventSourcing/IdempotentCommands/IdempotentCommandStreamDecorator.php
@@ -1,0 +1,94 @@
+<?php
+namespace Broadway\EventSourcing\IdempotentCommands;
+
+use Broadway\Domain\DomainEventStream;
+use Broadway\Domain\DomainEventStreamInterface;
+use Broadway\Domain\DomainMessage;
+use Broadway\Domain\Metadata;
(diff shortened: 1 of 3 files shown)


reenl:
Hi we had an equal issue, we resolved this by setting a flag on the aggregate when the event was being applied.

Note that this solution uses twice the resources when writing events (beccause it needs to load the events twice), and it is note compatible with snapshotting solutions. It is also not race condition save, because the first load (when checking commandId) might not return the same result as the second load (when building the aggregate). Sorry if this sounds like bashing the code, I actually like the developer interaction for this solution.

src/Broadway/EventSourcing/IdempotentCommands/IdempotentCommandStreamDecorator.php
```diff
@@ -0,0 +1,94 @@
...
+     *
+     * @return string|null
+     */
+    private function getCommandId(DomainMessage $domainMessage)
+    {
+        $metaData = $domainMessage->getMetadata()
+                                  ->serialize();
+
+        // @see http://stackoverflow.com/questions/4260086/php-how-to-use-array-filter-to-filter-array-keys
+        $commandId = array_intersect_key($metaData, array_flip([$this->metaDataKey]));
```
reenl: Why not simply use array_key_exists?

PR #302, opened 2017-02-26
load events starting from a given playhead

othillo:
This paves the way for snapshotting support, where you want to load only the events recorded since last snapshot.

Borrowed heavily from @kimlai's PR: https://github.com/broadway/broadway/pull/147

see https://github.com/broadway/broadway/issues/111

### DIFF
--- a/src/Broadway/EventStore/InMemoryEventStore.php
+++ b/src/Broadway/EventStore/InMemoryEventStore.php
@@ -44,6 +44,29 @@ class InMemoryEventStore implements EventStoreInterface, EventStoreManagementInt
     /**
      * {@inheritDoc}
      */
+    public function loadFromPlayhead($id, $playhead)
+    {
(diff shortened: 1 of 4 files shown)


src/Broadway/EventStore/InMemoryEventStore.php
```diff
@@ -44,6 +44,29 @@ public function load($id)
     /**
      * {@inheritDoc}
      */
+    public function loadFromPlayhead($id, $playhead)
+    {
+        $id = (string) $id;
+
+        if (!isset($this->events[$id])) {
```
wjzijderveld: Shouldn't this throw an exception? At least when given playhead is bigger then 0?

wjzijderveld: Hmm, more complicated then it seemed

PR #310, opened 2017-03-09
Mark some classes with `final`

unkind:
I suggest to lock some classes from inheritance before releasing 1.0.

Motivation:
* http://verraes.net/2014/05/final-classes-in-php/
* https://ocramius.github.io/blog/when-to-declare-classes-final/

In short, it is pain-free to mark as `final`:
* value objects; you don't need to mock/proxy them anyway: just create new instance of VO;
* classes with 1-to-1 interface: you always can create decorator with new behavior.

It makes easier to keep BC for your classes and forces to use composition instead of inheritance.


wjzijderveld:
Thanks @unkind ! Sorry for the delay 😃

PR #311, opened 2017-05-17
Add consistency to the upgrade guide

NoelDavies:
I've just added a hyphen but it's because i'm creating an automated script for our internal projects and this helped me parse it, and may help others.


othillo:
thank you, @NoelDavies !

are you considering open sourcing your script?

NoelDavies:
@othillo If there's a need for it, I don't see why not.

PR #312, opened 2017-05-29
Snapshotting

050media:
This PR adds Snapshotting functionality. As we need it for our own project, I've been reading over the comments and proposed implementations of Snapshotting for the framework (#111, #147).

A seperate repository was created for this over at https://github.com/broadway/snapshotting, but this PR introduced changes to some core classes.

I've attempted to introduce as few breaks as possible. The only ones I've had to do were changing the `AggregateFactory` interface and modifying `EventSourcedAggregate::initializeFromState`.

Currently, code duplication is introduced by the `SnapshottingEventSourcingRepository`. An alternative would be to extend the `EventSourcingRepository` and grant either protected access or use visibility scoping. It's somewhat messy in my opinion but, seeing as I was able to re-use the existing tests, I saw this as the lesser evil.

* To use Snapshotting, the `SnapshottingEventSourcingRepository` must be used.
* Aggregates passed to this repository must extends `EventSourcingAggregrate` and implement `Snapshotable`
* Aggregates can be converted to Snapshots using a `Snapshotter`
* When `Trigger` criteria are met, a Snapshot will be created after having persistend and published the events (Only an event-count based trigger is contained in this PR)
* Only an InMemory implementation of the `SnapshotStore` is contained in this PR
* New `Trigger` and `SnapshotStore` implementations will be able to be tested using the provided base classes

Credits to @othillo and @kimlai for the groundworks and reference material. I've been able to assimilate most of their work.


src/Broadway/Snapshotting/Snapshotable.php
```diff
@@ -0,0 +1,25 @@
...
+ *
+ *  (c) Qandidate.com <opensource@qandidate.com>
+ *
+ *  For the full copyright and license information, please view the LICENSE
+ *  file that was distributed with this source code.
+ */
+
+namespace Broadway\Snapshotting;
+
+interface Snapshotable
```
codeliner: Disclaimer: I'm not a Broadway user, so this comment only adds my 2c

Do you really need such an interface? It requires extra effort to implement and test this functionality. It adds a public method to a domain aggregate that should not be aware of snapshots because it is an infrastructure concern.

Also this introduces a second way of recreating internal state of the aggregate. Where you have only "event apply" methods before you now have an additional "apply snapshot" mechanism. This could lead to strange bugs if the aggregate behaves different with or without snapshot applied.

Why not simply serialize an aggregate using PHP build in functionality? This would be a real snapshot, not a user generated one.

src/Broadway/Snapshotting/Snapshotable.php
```diff
@@ -0,0 +1,25 @@
...
+{
+    /**
+     * @return array
```
codeliner: array what?
Can the array contain value objects?

src/Broadway/Snapshotting/SnapshotStore.php
```diff
@@ -0,0 +1,29 @@
...
+     * @param Snapshot $snapshot
+     */
+    public function save(Snapshot $snapshot);
+
+    /**
+     * @param mixed $id
+     *
+     * @return Snapshot
+     *
+     * @throws SnapshotNotFoundException
```
050media: I considered this as well but for whatever reason I didn't go through with it in the end. I do agree that it makes things a lot cleaner.

othillo:
I've created https://github.com/othillo/broadway-snapshotting as a separate repository specifically to not interfere (or BC break) with broadway/broadway.

You copying that code into this PR contradicts this.

The @broadway/broadway team is ok with promoting othillo/broadway-snapshotting to the broadway namespace to allow for people like yourself to help improve it.

How about we do that first and then see what can be improved?

050media:
I'm currently implementing the changes suggested by @codeliner which should resolve all BC-breaks. I'll cancel this PR. 

Do you want a new PR to go to your repository at https://github.com/othillo/broadway-snapshotting or https://github.com/broadway/snapshotting instead? Your own repository is using PHP7 while the rest of broadway is using PHP5

othillo:
@050media I've cherry-picked the commits in to this PR: https://github.com/broadway/snapshotting/pull/2

afaik I only dropped testing on PHP 5.6, we can reenable it if you want.

smitj:
It wasn't just testing, there's the `declare(strict_types=1);` code as well. Travis already picked it up it looks like. I'll keep an eye on the other repository. Thanks!

othillo:
yes, it's definitely PHP 7... 
I'm reluctant to support PHP 5.6 since it's no longer actively supported itself.

smitj:
I get what you're saying and I would normally agree with you. But it wouldn't really make sense though to have Broadway PHP5.6 compatible and having a component (from the same authors) which excludes this version.

Me quickly brushing over the code again does show that no new language features have been used, other than strict types and return values. Would it not be worth it to make it 5.6 compatible for now then killing off PHP5.6 support broadway-wide later?

(Disclaimer: Right now my organization is still stuck using 5.6, RIP us)

othillo:
I understand your situation. I've created a ticket for it: https://github.com/broadway/snapshotting/issues/3

If you could help out with backporting by sending a PR to https://github.com/broadway/snapshotting/pull/2 That would be great ;)

PR #319, opened 2017-07-31
Remove sprintf regression tests

reenl:
By default tests fail on my machine because the France locale is not available.
Assuming the en_US locale is available on more machines than fr_FR.

This commit will resolve the following issue

```
1) Broadway\Domain\DateTimeTest::it_creates_now with data set #1 ('fr_FR.UTF-8')
PHPUnit_Framework_Exception: The locale functionality is not implemented on your platform, the specified locale does not exist or the category name is invalid.

/.../broadway/test/Broadway/Domain/DateTimeTest.php:35
```

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -17,12 +17,6 @@ matrix:
     allow_failures:
         - php: hhvm
 
-addons:
-    apt:
-        packages:
-            - language-pack-fr
-            - language-pack-en
-
 before_install:
     # Disable XDebug speed up test execution. (Ignore failures if platform does not had the extension installed)
     - phpenv config-rm xdebug.ini || return 0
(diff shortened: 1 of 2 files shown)


othillo:
after looking at the original PR: https://github.com/broadway/broadway/pull/206/files and consulting @wjzijderveld I realised this test (actually the setLocale) is just a regression test for changing sprintf('%.6f') to sprintf('%.6F'). IMO we are just testing if we use sprintf correctly.

I'm OK with removing the setLocale and it's data provider altogether.

reenl:
Should be resolved

reenl:
Awesome side-effect, building takes 2 minutes less :)

wjzijderveld:
Thanks @reenl !

Issue #318, opened 2017-07-27
No way to interact with Metadata except for serialize


reenl:
I was just wondering if ArrayAccess should be added to Metadata, because there is no way to extract values other than serializing it.

othillo:
I like the fact that the Metadata object is immutable.
Maybe just add a get($key) method?

reenl:
Nice catch, I misread merge. 

Just to get the other options considered:

- Throwing on offsetSet can still enforce immutable, however in this case I prefer get()
- I dislike the need to use serialize on getting all values, I propose also adding values() or all() eventhough serialize would probably do the same.

othillo:
get() and all() sound fine (Symfony ParameterBag style).
PR is welcome ;)

othillo:
fixed by #321

PR #321, opened 2017-07-31
Provide access to Metadata values with `get()` and `all()`

reenl:
Before this commit it wasn't possible to read values from Metadata except when you serialized it.

Eventhough `serialize()` is functionally equal to `all()` it was decided at #318 to add it for semantics sake. In other words, `serialize()` must be used for persisting the Metadata, `all()` or `get()` must be used to act upon.

### DIFF
--- a/src/Broadway/Domain/Metadata.php
+++ b/src/Broadway/Domain/Metadata.php
@@ -29,6 +29,19 @@ class Metadata implements Serializable
     }
 
     /**
+     * Helper method to construct an instance containing the key and value.
+     *
+     * @param mixed $key
+     * @param mixed $value
+     *
+     * @return Metadata
+     */
+    public static function kv($key, $value)
+    {
+        return new Metadata([$key => $value]);
+    }
+
+    /**
      * Merges the values of this and the other instance.
      *
      * @param Metadata $otherMetadata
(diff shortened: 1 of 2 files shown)


src/Broadway/Domain/Metadata.php
```diff
@@ -43,24 +56,37 @@ public function merge(Metadata $otherMetadata)
...
      *
-     * @param mixed $key
-     * @param mixed $value
+     * @param string $key
      *
-     * @return Metadata
+     * @return mixed
      */
-    public static function kv($key, $value)
+    public function get($key)
```
othillo: instead of returning null should we return a specified default value like in ParameterBag? https://github.com/symfony/http-foundation/blob/v3.3.6/ParameterBag.php#L86-L89

It's more functional that way IMO.

reenl: I considered this however, the code of Symfony predates the `?:` operator to my knowledge, I would suggest writing the consumer like this:

```
$valueForX = $metadata->get('x') ?: 'default';
```

Problem gets more serious with fallbacks:
```
$valueForAbcWithFallbackToXyz = $metadata->get('abc') ?: $metadata->get('xyz');
// instead of 
$valueForAbcWithFallbackToXyz = $metadata->get('abc', $metadata->get('xyz'));
```

Let me know what you think, I would happily add the Symfony fallback behaviour if you still prefer it.

pamil: Using `?:` or `??` (if using PHP7) encourages lazy evaluation, whereas passing default value to the function requires evaluating it beforehand.

It also prevents nesting:
```php
$metadata->get('foo', $metadata->get('bar', $metadata->get('baz')));

$metadata->get('foo') ?? $metadata->get('bar') ?? $metadata->get('baz');
```

reenl: Did we just race condition ourselves into an answer? :) @pamil

othillo: thanks for your input, @pamil, @reenl !
Let's leave it as is and use PHP7 :)

wjzijderveld:
Thanks @reenl 
And thanks @pamil for you input :-)

PR #322, opened 2017-08-02
[POC|WIP] Ability to detect event sequence

reenl:
Hi,

I was playing around to find a structural way of being able to guarantee an event message is delivered in order and only once when applying them asynchronously. However `DomainMessage` does not contain the information required to create such a system.

Therefore I added the sequenceId to the DomainMessage. It cannot be added to the constructor, because DBAL would only have the sequenceId after inserting the DomainMessage.

Before merging I would still need to write tests for this. 

I have the following questions:

1) Does the broadway team like the way sequence is implemented in `DomainMessage`
2) Should we use the visitEvents construction, or create a separate method drivers are already forced to be ["complex"](https://github.com/broadway/event-store-dbal/blob/master/src/DBALEventStore.php#L294) (Separate method `EventStoreManagement::replayEventsBetween` in this PR)

Please let me know if the idea is clear, otherwise I can share the code I planned for a future PR for this to maybe make more sense.

Related docs:
- http://www.axonframework.org/apidocs/2.0/org/axonframework/domain/DomainEventMessage.html#getSequenceNumber()
- https://docs.axonframework.org/v/3.0/part3/event-processing.html#asynchronous-event-processing

Edit:
After discussing with wjzijderveld it seems I misinterpreted the use of sequence. Going to have a good night sleep over this :).

### DIFF
--- a/src/Broadway/Domain/DomainMessage.php
+++ b/src/Broadway/Domain/DomainMessage.php
@@ -34,7 +34,7 @@ final class DomainMessage
     /**
      * @var string
      */
-    private $id;
+    private $aggregateId;
 
     /**
      * @var DateTime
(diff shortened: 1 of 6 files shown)


reenl:
In retrospect, and after a little discussion yesterday with wjzijderveld, I will be closing this PR. Adding a global sequenceId would result in issues in the future (sharding for example). With a little workaround I can most likely implement this by tracking the playhead of aggregates.

PR #323, opened 2017-08-03
DomainMessage no longer uses inheritDoc

reenl:
The DomainMessageInterface was dropped, outdating the docblocks


wjzijderveld:
Indeed, good catch. Thanks!

PR #324, opened 2017-08-04
In memory repository can find by value object

francescotrucchia:
With this patch findBy method can find also on getter method that return a value object

### DIFF
--- a/src/Broadway/ReadModel/InMemory/InMemoryRepository.php
+++ b/src/Broadway/ReadModel/InMemory/InMemoryRepository.php
@@ -29,7 +29,7 @@ final class InMemoryRepository implements Repository, Transferable
      */
     public function save(Identifiable $model)
     {
-        $this->data[(string) $model->getId()] = $model;
+        $this->data[(string)$model->getId()] = $model;
     }
 
     /**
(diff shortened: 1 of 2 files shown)


francescotrucchia:
@othillo any update for this PR?

wjzijderveld:
Would you mind adding a test for this? https://github.com/broadway/broadway/blob/master/test/Broadway/ReadModel/RepositoryTestCase.php

reenl:
Not sure about this but a php object that doesn't have a __toString method would generate a catchable fatal error with this change:

```
❯ php -a
Interactive mode enabled

php > class A {}
php > $a = new A();
php > echo (string) $a;
PHP Catchable fatal error:  Object of class A could not be converted to string in php shell code on line 1
PHP Stack trace:
PHP   1. {main}() php shell code:0
```

francescotrucchia:
@reenl I fix the behaviour. 
@wjzijderveld I add the test.
Thanks for feedback

francescotrucchia:
Hi guys, any update for this PR? Thanks

othillo:
@francescotrucchia can you rebase this, please?

PR #325, opened 2017-08-19
Add base test case and scenario to test processor

francescotrucchia:
Scenario to test processor

### DIFF
--- a/src/Broadway/CommandHandling/ClosureCommandHandler.php
+++ b/src/Broadway/CommandHandling/ClosureCommandHandler.php
@@ -9,6 +9,10 @@
  * file that was distributed with this source code.
  */
 
+/*
+ * @author Francesco Trucchia <francesco@trucchia.it>
+ */
+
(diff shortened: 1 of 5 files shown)


francescotrucchia:
Hi guys, any chance to receive a review for this PR? Thanks

reenl:
Please correct me if I'm wrong but the way I understood the processor was that its purpose was to do something with events, like sending e-mails or calling a webhook. Ofcourse you can inject the CommandBus but it is not requirement. For Event -> Command flow an attempt has been made to create a Saga library, but afaik it is experimental.

I'm not a maintainer so please wait for them for an official response :)

othillo:
I agree with @reenl here. I'm not comfortable with injecting a command bus into a processor.
To me a processor is just an absolute minimal implementation of EventListener. You can do whatever you want which does mean you have to write the tests yourself.

othillo:
Thanks, @francescotrucchia! But I don't think this should be merged. See the earlier remarks.

PR #326, opened 2017-08-19
Add chance to append an event stream of events with different id

francescotrucchia:
Sometimes it could be useful to append on event store in a single transaction a domain stream of events with different id. DBALEventStore don't depend on append $id.


wjzijderveld:
I don't think this is a good idea to do. It's generally accepted that an aggregate [is a transaction boundary](http://cqrs.nu/Faq/aggregates).
Do you have a specific use-case where you would like to use this?

PR #327, opened 2017-09-07
Moved generic test classes to main namespace

robinvdvleuten:
To have some "generic" test cases that can be used in the elasticsearch and mongodb implementations, it would be nice to have this testcase classes moved to the src directory.

### DIFF
--- a/src/Broadway/ReadModel/Testing/RepositoryTestCase.php
+++ b/src/Broadway/ReadModel/Testing/RepositoryTestCase.php
@@ -9,19 +9,25 @@
  * file that was distributed with this source code.
  */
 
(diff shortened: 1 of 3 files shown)


othillo:
While I'm OK with this (we're about to break BC in version 2.0 anyway), it makes sense to apply this to all base test classes.

As a workaround you can use this: https://github.com/broadway/read-model-elasticsearch/blob/master/composer.json#L47

robinvdvleuten:
@othillo sounds indeed like a good workaround for now

othillo:
the workaround is used by

- [ ] https://github.com/broadway/event-store-dbal/blob/master/composer.json#L50

PR #328, opened 2017-10-12
Added reflection serializer

abachmann:
In order to decouple events from the broadway framework I implemented a reflection serializer.


othillo:
thank you, @abachmann!

PR #329, opened 2017-10-24
apply coding standards

othillo:
We use Symfony coding standards since https://github.com/broadway/broadway/pull/320.
This PR applies them.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,2 +1,3 @@
 /vendor/
 composer.lock
+.php_cs.cache
--- a/.travis.yml
+++ b/.travis.yml
@@ -27,7 +27,8 @@ install:
     - composer info -i
 
 script:
-  - phpunit --exclude-group=none
+  - vendor/bin/phpunit --exclude-group=none
+  - vendor/bin/php-cs-fixer fix --config=".php_cs" --dry-run --no-interaction --allow-risky=yes --diff --verbose
 
 branches:
   only:
--- a/composer.json
+++ b/composer.json
@@ -38,7 +38,8 @@
         "instaclick/base-test-bundle": "~0.5",
         "monolog/monolog": "~1.8",
         "phpunit/phpunit": "^4.8",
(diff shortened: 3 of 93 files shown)


msvrtan:
Hey @othillo , 

I see you are running a major code standards apply, would you consider adding this rules? 

```
        '@Symfony' => true,
        'array_syntax' => ['syntax' => 'short'],
        'binary_operator_spaces' => [
            'align_double_arrow' => true,
            'align_equals' => true
        ],
        'ordered_imports' => true,
        'phpdoc_order' => true,
```

To explain the suggestions:

`        'array_syntax' => ['syntax' => 'short'], `
I guess since you added strict types declarations that 5.x is not supported any more so this should not be an issue as well and helps readability...

```
        'binary_operator_spaces' => [
            'align_double_arrow' => true,
            'align_equals' => true
        ],
```
I think that having equals and double arrows alligned helps the code readbility a lot IMO

`        'ordered_imports' => true,`
This would sort the imports in alphabetical order, while broadway doesnt have that much imports per file, might not be needed but not sure anyone would mind (especially OCD ppl like me might be gratefull :) )


`        'phpdoc_order' => true,`
Same thing as last one, not sure it's needed but might help with future contributors to avoid (can you please reorder the docblocks)

othillo:
hey Miro!

you're talking to a fellow OCD person here ;)
I understand and personally like your suggestions. In fact they were part of our previous but unofficial coding standards.

The thing is: I don't want to bother contributors with coding standards, especially not when they are not standard. Which is my second point: I want them to be a simple as possible.
That's why we chose the Symfony standard (with strict types). And to enforce it on CI this PR is necessary.

asm89:
👍

msvrtan:
I understand completely, I was just expecting that adding those rules will not create any issues since to be up-to Symfony standard you would be running php-cs-fixer too (at least I would since I dont even try to manually guess the correct details :) )

PR #330, opened 2017-10-24
drop HHVM testing on Travis CI

othillo:


Xerkus:
https://hhvm.com/blog/2017/09/18/the-future-of-hhvm.html
> PHP7 is charting a new course away from PHP5, and we want to do the same, via a renewed focus on Hack. Consequently, HHVM will not aim to target PHP7. The HHVM team believes that we have a clear path toward making Hack a fantastic language for web development, untethered from its PHP origins. We’d do ourselves and our users a disservice by positioning HHVM as an uncommon, less well-documented, less compatible PHP7 runtime.

I hear that HHVM as an alternative php runtime does not bring significant performance increase, if any, compared to php 7. That and divergent goals of the two makes it impractical to maintain compatibility, effort better be served by utilizing more of php 7 capabilities.

othillo:
https://symfony.com/blog/symfony-4-end-of-hhvm-support

PR #331, opened 2017-10-28
PHPUnit 6

abachmann:


composer.json
```diff
@@ -37,7 +37,7 @@
     "require-dev": {
         "instaclick/base-test-bundle": "~0.5",
         "monolog/monolog": "~1.8",
-        "phpunit/phpunit": "^4.8",
+        "phpunit/phpunit": "^6.4",
```
othillo: please use ^6.0

test/Broadway/CommandHandling/ClosureCommandHandlerTest.php
```diff
@@ -35,21 +33,23 @@ public function it_delegates_command_to_proper_handle_function()
 
     /**
      * @test
+     *
+     * @expectedException \Broadway\CommandHandling\Exception\CommandNotAnObjectException
      */
     public function it_throws_when_handling_a_non_object_command()
     {
         $commandHandler = new ClosureCommandHandler();
-        $this->setExpectedException(CommandNotAnObjectException::class);
```
othillo: please use expectException(), see https://thephp.cc/news/2016/02/questioning-phpunit-best-practices

othillo:
FYI: this PW has merge conflicts with https://github.com/broadway/broadway/pull/329 which is planned to be merged first.

abachmann:
@othillo Thanks for the link! I alwya used annotations before.

Once the PR #329 is merged I revert my changes to the `ReflectionSerializer`.

abachmann:
I resolved the conflicts

test/Broadway/Auditing/CommandLoggerTest.php
```diff
@@ -17,12 +17,27 @@
...
+     */
     private $command;
+
+    /**
+     * @var CommandLogger
+     */
     private $commandAuditLogger;
+
+    /**
+     * @var CommandSerializer|\PHPUnit_Framework_MockObject_MockObject
```
othillo: just CommandSerializer?

test/Broadway/Auditing/CommandLoggerTest.php
```diff
@@ -17,12 +17,27 @@
 
 class CommandLoggerTest extends TestCase
 {
+    /**
+     * @var LoggerInterface
```
othillo: TraceableLogger

test/Broadway/Auditing/NullByteCommandSerializerTest.php
```diff
@@ -17,10 +17,17 @@
 
 class NullByteCommandSerializerTest extends TestCase
 {
+    /**
+     * @var CommandSerializer
```
othillo: NullByteCommandSerializer

test/Broadway/CommandHandling/EventDispatchingCommandBusTest.php
```diff
@@ -13,29 +13,41 @@
 
 namespace Broadway\CommandHandling;
 
+use Broadway\EventDispatcher\EventDispatcher;
 use Broadway\TestCase;
 
 class EventDispatchingCommandBusTest extends TestCase
 {
+    /**
+     * @var CommandBus|\PHPUnit_Framework_MockObject_MockObject
```
othillo: I'm not a big fan of these annotations because they are not automatically validated.
In case of mocking I think it's best to just specify the base class not the mock.

test/Broadway/CommandHandling/SimpleCommandBusTest.php
```diff
@@ -100,9 +103,13 @@ public function it_should_still_handle_commands_after_exception()
         $this->commandBus->dispatch($command2);
     }
 
+    /**
+     * @param $expectedCommand
+     * @return SimpleCommandHandler|\PHPUnit_Framework_MockObject_MockObject
```
othillo: in this case it only returns the mock.

othillo: can you add it as a return type hint? that's better than doc blocks.

test/Broadway/EventHandling/SimpleEventBusTest.php
```diff
@@ -141,9 +144,12 @@ public function it_should_still_publish_events_after_exception()
         $this->eventBus->publish($domainEventStream2);
     }
 
+    /**
+     * @return EventListener|\PHPUnit_Framework_MockObject_MockObject
```
othillo: same comment as above

othillo:
@abachmann thanks again! The expectException calls are a definite improvement.
I have some remarks on the type descriptions in the docblocks and some return types that can be added. Like I said, not a big fan of docblocks because they are error prone and need manual maintenance. But if they are correct I don't mind merging them. Also I still see some conflicts.

test/Broadway/CommandHandling/EventDispatchingCommandBusTest.php
abachmann: Without this annotation the code completion of an IDE is not working. That's why I added it. But I can remove it.

othillo: ok, that's a fair reason :)
you can leave it.

test/Broadway/CommandHandling/SimpleCommandBusTest.php
abachmann: Ok, can we set "SimpleCommandHandler" as return type? Because this is what it should be in the end.

othillo: yes, please

othillo:
thank you, @abachmann !

PR #332, opened 2017-11-08
Strip leading slash to allow testing repository with class keyword

rgeraads:
When testing a repository that has the `::class` keyword instead of a string representation, a test failed because that expected a leading slash.

The fix allows for both FQN with and without leading slash so it is BC

Thanks @reenl for helping me out!

### DIFF
--- a/test/Broadway/EventSourcing/AbstractEventSourcingRepositoryTest.php
+++ b/test/Broadway/EventSourcing/AbstractEventSourcingRepositoryTest.php
@@ -146,7 +146,7 @@ abstract class AbstractEventSourcingRepositoryTest extends TestCase
         $lastCall = $this->eventStreamDecorator->getLastCall();
 
         $this->assertEquals($aggregate->getAggregateRootId(), $lastCall['aggregateIdentifier']);
-        $this->assertEquals(get_class($aggregate), $this->stripLeadingSlash($lastCall['aggregateType']));
+        $this->assertInstanceOf($lastCall['aggregateType'], $aggregate);
 
         $events = iterator_to_array($lastCall['eventStream']);
         $this->assertCount(1, $events);
--- a/test/Broadway/EventSourcing/AbstractEventSourcingRepositoryTest.php
+++ b/test/Broadway/EventSourcing/AbstractEventSourcingRepositoryTest.php
@@ -146,7 +146,7 @@ abstract class AbstractEventSourcingRepositoryTest extends TestCase
         $lastCall = $this->eventStreamDecorator->getLastCall();
 
         $this->assertEquals($aggregate->getAggregateRootId(), $lastCall['aggregateIdentifier']);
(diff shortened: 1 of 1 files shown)


rgeraads:
```
0.15s$ phpunit --exclude-group=none
PHPUnit 6.3.0 by Sebastian Bergmann and contributors.
Time: 87 ms, Memory: 10.00MB
No tests executed!
```
😟

reenl:
Thanks for the mention! Seems master is still running phpunit 4.8, however that one crashes because it is not able to mock return types.

othillo:
check this PR: https://github.com/broadway/broadway/pull/331

rgeraads:
I can rebase my changes to that PR if you want

test/Broadway/EventSourcing/AbstractEventSourcingRepositoryTest.php
```diff
@@ -146,7 +146,7 @@ public function it_calls_the_event_stream_decorators_with_the_correct_arguments(
         $lastCall = $this->eventStreamDecorator->getLastCall();
 
         $this->assertEquals($aggregate->getAggregateRootId(), $lastCall['aggregateIdentifier']);
-        $this->assertEquals('\\' . get_class($aggregate), $lastCall['aggregateType']);
+        $this->assertEquals(get_class($aggregate), $this->stripLeadingSlash($lastCall['aggregateType']));
```
othillo: so you're saying https://github.com/broadway/broadway/blob/master/src/Broadway/EventSourcing/EventSourcingRepository.php#L51 allows classnames with slashes and without?
This is validated by assertExtendsEventSourcedAggregateRoot  -> Assert::subclassOf which in turn calls PHP's is_subclass_of function in https://github.com/beberlei/assert/blob/master/lib/Assert/Assertion.php#L1306

wouldn't it be easier to just use assertTrue(is_subclass_of($aggregate, $lastCall['aggregateType'])) ?
or maybe even assertInstanceOf($lastCall['aggregateType'], $aggregate) ?

othillo: Would you mind adding a test for both situations for sake of regression testing?

rgeraads: I don't understand how, since this already is a test

othillo: You mentioned it breaks when creating a repository using aggregate::class notation and not when using the string FQDN representation. I would like to see a test for both conditions. 

But I've tried it manually by modifying https://github.com/broadway/broadway/blob/master/test/Broadway/EventSourcing/EventSourcingRepositoryTest.php#L29 and your change works, so I'm ok with it.

othillo:
thank you, @rgeraads !

PR #336, opened 2017-11-13
PHPUnit 6 fixes

abachmann:
All Broadway test classes are still inheriting from `PHPUnit_Framework_TestCase`. This breaks all project-related tests - sorry for that! This PR will fix it.

I create a new PR to add tests to avoid these issues in the future.


othillo:
thank you, @abachmann !

PR #340, opened 2017-11-13
PHP 7.1

othillo:
issues found using https://github.com/thecodingmachine/phpstan-strict-rules

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -9,7 +9,6 @@ cache:
 
 matrix:
     include:
-        - php: 7.0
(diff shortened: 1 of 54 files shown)


wjzijderveld:
Is upgrading a minimum version a BC break? I guess not right? As it can't break things, you just won't be able to upgrade?

othillo:
correct! it is explained here: https://www.doctrine-project.org/2017/07/25/php-7.1-requirement-and-composer.html

PR #342, opened 2017-12-05
test: strictly encode & decode event payload

rgeraads:
I noticed this test passed while deserialize() did not properly return the same object.
Typehinted parameters were ignored because everything is handled in the same request.
image

Explicitly encoding/decoding the payload detects this bug

### DIFF
--- a/src/Broadway/Serializer/Testing/SerializableEventTestCase.php
+++ b/src/Broadway/Serializer/Testing/SerializableEventTestCase.php
@@ -39,6 +39,8 @@ abstract class SerializableEventTestCase extends TestCase
         $event = $this->createEvent();
 
         $serialized = $serializer->serialize($event);
+        $serialized['payload'] = json_decode(json_encode($serialized['payload']), true);
+
         $deserialized = $serializer->deserialize($serialized);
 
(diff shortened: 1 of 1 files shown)


othillo:
I'm not following this one. Can you write a test demonstrating this issue?

rgeraads:
Will do! you're gonna love this

rgeraads:
@othillo here is a test: https://3v4l.org/6AMZ5

the 1st var_dump passes, but it should fail because the correct code should be:
```
    public static function deserialize(array $data)
    {
        return new self(new Id($data['id']));
    }
```

othillo:
OK, so in your case even though serializing and deserializing work according to the interface the result may not survive a json_encode and json_decode. (Normalization works but encoding doesn't). 

The serializers in Broadway are responsible for normalizing, not encoding. Therefore, encoding is not tested in the SerializableEventTestCase. I suggest adding a test in your application to test both depending on the encoding you use. Or you can use the ReflectionSerializer to make sure nested objects are converted to arrays with primitive types.

rgeraads:
Ok, I see! I'll close this PR since encoding/decoding should not be part of broadway.

PR #343, opened 2017-12-05
Simplified type handling in event sourcing repository

othillo:


src/Broadway/EventSourcing/EventSourcingRepository.php
```diff
@@ -87,7 +87,7 @@ public function save(AggregateRoot $aggregate)
 
     private function decorateForWrite(AggregateRoot $aggregate, DomainEventStream $eventStream): DomainEventStream
     {
-        $aggregateType = $this->getType();
+        $aggregateType = get_class($aggregate);
```
wjzijderveld: I think that technically this is a BC break? As `$aggregate` can be an extended class from `$this->aggregateClass`. With this change the decorators get the actual class... so it's also a bugfix :stuck_out_tongue: ?

othillo: To me it's just a bugfix. Indeed, the decorators didn't get the actual (extended) class before this change.

Also it's safer to rely on get_class() and ::class than manually specifying the FQCN.

Do you think people rely on the old behaviour?

wjzijderveld: No, I doubt it. :+1:

PR #344, opened 2017-12-15
Bugfixes in event-sourced-domain-with-tests

renedekat:
Fixed missing return types required in strict mode. 

Fixed bug in Scenario->getEvents method: array_map only accepts an array and not an iterator. Using iterator_to_array() now.

Renamed InvitationTest class name to match filename.

### DIFF
--- a/examples/event-sourced-domain-with-tests/Invites.php
+++ b/examples/event-sourced-domain-with-tests/Invites.php
@@ -34,7 +34,7 @@ class Invitation extends Broadway\EventSourcing\EventSourcedAggregateRoot
      *
      * {@inheritdoc}
      */
-    public function getAggregateRootId()
+    public function getAggregateRootId() : string
     {
         return $this->invitationId;
     }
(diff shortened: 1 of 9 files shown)


examples/event-sourced-domain-with-tests/Invites.php
```diff
@@ -34,7 +34,7 @@ public static function invite($invitationId, $name)
      *
      * {@inheritdoc}
      */
-    public function getAggregateRootId()
+    public function getAggregateRootId() : string
```
othillo: apparently the CS fixers fails. Can you remove the space before the colon? "public function getAggregateRootId(): string"

examples/event-sourced-domain-with-tests/InvitesTest.php
```diff
@@ -23,7 +23,7 @@ public function setUp()
         $this->generator = new Broadway\UuidGenerator\Rfc4122\Version4Generator();
     }
 
-    protected function getAggregateRootClass()
+    protected function getAggregateRootClass():string
```
othillo: protected function getAggregateRootClass(): string

sorry, the CS fixer is picky

examples/event-sourced-domain-with-tests/Invites.php
renedekat: Done, as well as added the examples folder to phpunit.xml.dist

othillo:
Thank you, @renedekat !

renedekat:
Now I just have to figure how to use it in a project :) I'm new to CQRS and Event Sourcing.

PR #346, opened 2017-12-16
improve the documentation

othillo:


docs/examples.md
```diff
@@ -0,0 +1,9 @@
+## Examples
+
+Examples can be found in the [`examples/`][examples] directory. Most of the
+examples focus on showing how one of the components works. There is also a more
+[deliberate example][example] using several components and showing how you can
+test your event sourced model.
+
+[examples]: examples/
```
wjzijderveld: Shouldn't this be `/examples/` or `../examples/`? At least in the preview it now links to `docs/examples`

docs/integrations.md
```diff
@@ -0,0 +1,10 @@
+# Integrations
+
+- The broadway project ships with a [bundle] to use with a Symfony application.
```
wjzijderveld: it's not shipped anymore: `There is a [bundle] available to use with your Symfony application`

PR #349, opened 2018-01-15
enforce license

othillo:


mdwheele:
Any reason to use docheader over "header_comment" from php-cs-fixer?

othillo:
I wasn't aware of the header_comment rule. Tnx for pointing it out. It makes more sense to use that instead of introducing another tool.

othillo:
unfortunately there is an open issue with header_comment and declare_strict_types: https://github.com/FriendsOfPHP/PHP-CS-Fixer/issues/2590

wjzijderveld:
Is this still WIP?
I suggest we merge this? We can always change to a php-cs-fixer solution in the future.

PR #355, opened 2018-08-01
Update beberlei/assert to ~3.0

Lutacon:
Hello,

We are using broadway but also want to have the new features from beberlei/assert. Just basically the fix `AssertionFailedException` 👀 

On 3.0 they are dropping support of PHP < 7.0

https://github.com/beberlei/assert/releases

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -4,7 +4,7 @@
     "keywords": ["cqrs", "event sourcing", "domain-driven design", "ddd"],
     "license": "MIT",
     "require": {
-        "beberlei/assert": "~2.0",
+        "beberlei/assert": "~3.0",
         "broadway/uuid-generator": "~0.4.0",
         "php": ">=7.1"
     },


composer.json
```diff
@@ -4,7 +4,7 @@
     "keywords": ["cqrs", "event sourcing", "domain-driven design", "ddd"],
     "license": "MIT",
     "require": {
-        "beberlei/assert": "~2.0",
+        "beberlei/assert": "~3.0",
```
ruudk: Why not allow both `~2.0 || ~3.0`. There is no real reason to only required 3.0.

ruudk:
I added a comment to also allow 2.0. Would be nice it this can be merged as it's currently blocking us from upgrading to PHP 7.3 (since url regex is broken in Assert 2.0)

composer.json
wjzijderveld: There is not a hard reason, but also doesn't add a lot of value imo. 
We require PHP7, beberlei/assert 3.0 is released for PHP7, so I see no reason not make that the requirement.

wjzijderveld:
Thanks @Lutacon !

ruudk:
Great, will it be tagged as a new release?

PR #356, opened 2018-08-07
Add examples for creating read models

fritsjanb:
I added two examples for creating readmodels, one example using ReadModelScenario/ReadModelRepository and one example that does not use the (opiniated) provided functionaly.

Solves broadway#345


othillo:
thank you, @fritsjanb !

PR #358, opened 2018-08-07
Simple command bus - Manage Throwable exceptions

felixcarpena:
# Goal:
This PR aims to allow simple command bus to manage all kind of exceptions, using "finally" as you are already using for the event bus implementation, in order to flag it as isDispatching false either if it went ok or ko (exception thrown)
Related issue: https://github.com/broadway/broadway/issues/357


wjzijderveld:
Thanks @felixcarpena !

PR #359, opened 2018-08-12
Change the order of event applying in aggregate from root-to-leaf to leaf-to-root

Ormin:
This PR aims to reverse the order of applying the event over an aggregate - instead of applying it top-down, event should be applied from the leafs and work towards the root.

The reason behind that is that when applying an event on an entity within the aggregate, we cannot rely on our child entities ( because they didn't apply the event yet ).

Let me introduce a minimalistic example - Imagine a classic shopping basket with a twist - if basket's contents exceed 100 USD, we mark a basket as important ( Management cares about big baskets! ) and we allow the basket holder to do something more. We also allow to mark a basket as important explicitly ( because why not and it's just an example :) ). And we also are able to apply a discount to a product within a basket. Let's write it out ( only relevant parts are here ):

`

class Product extends SimpleEventSourcedEntity {

      private $price;

      public function applyDiscountGranted(DiscountGranted $event) {
              $this->price = (100 - $event->discount()) / 100 * $this->price;
      }
}

class Basket extends EventSourcedAggregateRoot {
    
     private $products;

     private $important;
     
    public function applyDiscountGranted(DiscountGranted $event){
     {
           if($this->products->sum() >= 100) {
                   $this->important = true;
           }
     }

     public function becomeAnImportantBasket()
     {
          //No domain-specific rules, anyone can explicitly ask to become an important basket. :)
          $this->apply(new BecomeImportant());
     }

    public function applyBecomeImportant(BecomeImportant $event) {
        $this->important = true;
    }

     public function doSomethingSpecial()
     {
            if($this->important) {
                 //Do something special with our important basket
            }         
     }
    
}
`

Now, the problem is that this aggregate would malfunction under Broadway, because when doing our comparison check within Basket::applyDiscountGranted() function we'll have a wrong sum of products' prices, and this is because Basket::applyDiscountGranted() is called before Product::applyDiscountGranted().

Of course, one could argue that we shouldn't modify state based on child entities' state within event applying routines, so let's try that for a moment:

`

class Product extends SimpleEventSourcedEntity {

      private $price;

      public function applyDiscountGranted(DiscountGranted $event) {
              $this->price = (100 - $event->discount()) / 100 * $this->price;
      }
}

class Basket extends EventSourcedAggregateRoot {
    
     private $products;

     private $important;

     public function becomeAnImportantBasket()
     {
          //No domain-specific rules, anyone can explicitly ask to become an important basket. :)
          $this->apply(new BecomeImportant());
     }

    public function applyBecomeImportant(BecomeImportant $event) {
        $this->important = true;
    }

     public function doSomethingSpecial()
     {
            if($this->isImportant()) {
                 //Do something special with our important basket
            }         
     }

     private function isImportant()
     {
           return ($this->products->sum() >= 100 || $this->important);
     }
    
}
`

However, this approach has a subtle problem. Imagine that business rules change ( they always change! ) and the "important basket threshold" changes to 200 USD. However, when we make the change, the problem becomes evident - playing back events from the past will break, because they will now all adhere to the new threshold, and surely events that happened in the past ought not to change state based on current rules. This is a classic events' versioning problem, one that can be easily solved by versioning events, but to do so, we need to version it within the applying function, but we cannot, because we took the logic out of it.

Another approach one could take is to save this information within the event, but this obviously clutters the event with the information that isn't relevant to it, and hence I discard it.

If you see any fallacy in my logic behind this PR or think I didn't understand something well, make sure to point it out, as this PR is also meant to be a discussion for/against rather than mere code change.

### DIFF
--- a/examples/event-sourced-child-entity/Parts.php
+++ b/examples/event-sourced-child-entity/Parts.php
@@ -66,11 +66,7 @@ class Part extends Broadway\EventSourcing\EventSourcedAggregateRoot
 
     protected function getChildEntities(): array
     {
-        // Since the aggregate root always handles the events first we can rely
-        // on $this->manufacturer being set by the time the child entities are
-        // requested *provided* PartWasManufacturedEvent is the first event in
-        // the event stream.
-        return [$this->manufacturer];
+        return ($this->manufacturer !== null )  ? [$this->manufacturer] : []; 
     }
 }
 
--- a/examples/event-sourced-child-entity/Parts.php
+++ b/examples/event-sourced-child-entity/Parts.php
@@ -58,6 +58,7 @@ class Part extends Broadway\EventSourcing\EventSourcedAggregateRoot
         // the child entity is instantiated and returned by getChildEntities()
         // it can emit and apply events itself.
         $this->manufacturer = new Manufacturer(
+            $this,
             $event->partId,
             $event->manufacturerId,
             $event->manufacturerName
@@ -76,8 +77,9 @@ class Manufacturer extends Broadway\EventSourcing\SimpleEventSourcedEntity
     private $manufacturerId;
     private $manufacturerName;
 
-    public function __construct($partId, $manufacturerId, $manufacturerName)
+    public function __construct(Broadway\EventSourcing\EventSourcedAggregateRoot $aggregateRoot, $partId, $manufacturerId, $manufacturerName)
     {
+        parent::__construct($aggregateRoot);
         $this->partId = $partId;
         $this->manufacturerId = $manufacturerId;
         $this->manufacturerName = $manufacturerName;
--- a/examples/event-sourced-multiple-dyanmic-child-entities/JobSeekers.php
+++ b/examples/event-sourced-multiple-dyanmic-child-entities/JobSeekers.php
@@ -73,6 +73,7 @@ class JobSeeker extends Broadway\EventSourcing\EventSourcedAggregateRoot
     public function applyJobWasAddedToJobSeekerEvent(JobWasAddedToJobSeekerEvent $event)
     {
         $this->jobs[$event->jobId] = new Job(
+            $this,
             $event->jobSeekerId,
             $event->jobId,
             $event->title,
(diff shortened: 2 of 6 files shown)


Ormin:
Eh, does someone know how to fix those code comments? It's sort of broken :(

Ormin:
Fixed some of the unit tests' fails, however, there's still some fails due to a bug in broadway where trying to apply the very first aggregate event within the children entity will result in a fatal error. Will try to address that in another PR.

UlrichEckhardt:
You PR is a few months old, the question is timeless though. I'm relatively new to the topic and this framework, so bear with me...

Concerning the term `Product`, I'd rather distinguish between a `Product` and an `Order`. This is mostly a rename, but it makes things a bit clearer. Of course though, an `Order` then references a `Product` and a count. Then, the event giving a discount becomes a `GiveDiscountForOrder`. Further, the `Basket` aggregate listens to `GiveDiscountToOrder` events and updates its amount whenever one of its nested `Orders` changes. You could do so by overriding `handleRecursively()` and updating the amount after regular event processing.

Note: I'm actually not too happy with the term `Basket` even. Actually, this should be called `Order`, with different `Items` that reference a `Product` and how many, but that's just a suggestion and your use case may vary widely.

Concerning the importance of a `Basket`, it should be defined as a (read-only) property. It is true when either the amount exceeds 100 USD or when it sees an event "OperatorMarkedBasketAsImportant". In other words, your code is completely right, except (IMHO) for the naming of the event. Manually marking a basket as important is something different than a basket becoming important, although the former is one possible cause for the latter. As a general rule, name the events after what just happened.

Now, you ask how to apply a limit change. If that limit is part of your aggregate, then it can not change, simple as that. You can create a new aggregate with a different limit, but then you have two different ones. The important point is that the ability to replay is simply not given when the rules change! However, you could also create an event `OperatorChangedImportantBasketThreshold`. In that case, your aggregate has a default limit (100 USD) but also listens to this new event, pulling its limit from it. Changing the limit doesn't change the importance. As a dirty hack, you could also write code that applies different limits depending on record date of the event being applied.

One warning though: It seems that you want to emit an event whenever an order has become important. However, if you have multiple definitions of important (100 vs 200) via different aggregates, this wont work! Also, the point of the aggregate is to collect data from events to provide an speed-optimized, read-only representation of that data, not to act on that data (which would in turn emit events). Of course, an aggregate can emit events to a UI that is then refreshed, but that's a different kind of event.

PR #360, opened 2018-08-21
Query event store example

wjzijderveld:
The event-store-dbal example is a bit hacky, but only if you actually want to run it :innocent:


othillo:
thank you, @wjzijderveld !

PR #361, opened 2018-08-21
Add replay events feature

jorge07:
In some use cases you need to replay certain events.

Let's say a bug introduced an error in aggregateRootID `foo`, version `32` and `33`.
As events are immutable I want to be able to add manually an event, let's say version `32.1` and replay the events to recompose the history.

This PR introduce the replay feature, but doesn't change the playhead attribute to float.

### DIFF
--- a/src/Broadway/EventSourcing/EventSourcingRepository.php
+++ b/src/Broadway/EventSourcing/EventSourcingRepository.php
@@ -35,11 +35,9 @@ class EventSourcingRepository implements Repository
     private $aggregateFactory;
 
     /**
-     * @param EventStore             $eventStore
-     * @param EventBus               $eventBus
-     * @param string                 $aggregateClass
-     * @param AggregateFactory       $aggregateFactory
      * @param EventStreamDecorator[] $eventStreamDecorators
+     *
+     * @throws \Assert\AssertionFailedException
      */
     public function __construct(
         EventStore $eventStore,
@@ -73,6 +71,8 @@ class EventSourcingRepository implements Repository
 
     /**
      * {@inheritdoc}
+     *
+     * @throws \Assert\AssertionFailedException
      */
     public function save(AggregateRoot $aggregate): void
     {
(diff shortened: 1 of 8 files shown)


pascal08:
What about splitting the interfaces so that it wouldn't force you to implement those methods if you don't need the replay functionality?

jorge07:
ping @othillo ;)

pascal08:
Would be great to have this merged. :)

fritsjanb:
#363 shows how the current code base can already be used to replay a single aggregate.

pascal08:
1) I agree with @jorge07 that this MR proposes something different than what is shown in #363. I don't see how it is possible to replay up to a certain playhead without this MR.
2) If we want to replay large event stream we need a buffer mechanism. I'd be glad open a PR for this if we can agree on Broadway supporting this kind of replays.

jorge07:
@pascal08 this this PR you can paginate if needed

pascal08:
@jorge07 Ah, I see. That would make (2) pointless.

jorge07:
Any update on this?

UlrichEckhardt:
I'm not sure if you're still following this, but from a puristic eventsourcing point of view, your approach is flawed:
 * Firstly, you say you can't replay all events, but that should be considered a bug in your code and fixed.
 * Secondly, instead of changing the events (which should only record indisputable facts!) you should change the code that interprets these events.

Doing the above would allow you just set up a new deployment and replay all events with the changed interpretation. If you don't want that and the misuse of eventsourcing is okay for you, you could also go ahead and change the DB directly, no need to add code for that to Broadway, IMHO.

That said, being able to replay to a certain point is a useful addition. Also, not exhausting memory (in case that's the problem you're hitting) when replaying lots of events is also a useful addition. One approach to the latter is using `yield` instead of building arrays, so the objects don't have to be in-memory all time.

jorge07:
Hi @UlrichEckhardt 

1 - I can replay 1TB of events. But there's no need to do this. It's a waste of time and resources with a close to 0 business value and a big cost. I just need to replay from version X to Y to rebuild my history due to a bug. My system it's idempotent so there's no need to replay all the events at all.
2 - Events must be immutable. That's the 1 law and I agree 100% with it but must to be consistent too. And when in real life, after an issue, aren't I prefer mutate the data instead of patch multiple services with the cost it generates.

+1 for `yield` good catch

NikoGrano:
@jorge07 I just want inform you, I will be merging this change to my [fork](https://github.com/HouseHold/broadway), if thats okay for you. 

Cheers!

PR #363, opened 2018-09-13
Add example of how to replay events using EventStoreManagement and Criteria

fritsjanb:
Example of how to create a replayer that uses EventStoreManagement and Criteria to replay a single EventListener


othillo:
thank you, @fritsjanb !

PR #370, opened 2019-01-09
Improvements to serializer code

UlrichEckhardt:
I stumbled across a few glitches in the serializers that caused a bit of confusion while trying to integrate Broadway into one of our projects. For some of the bugs, I created according commits that address them, both concerning the tests and the actual code. However, there's one problem left, meaning that this pull request should not yet be merged.

The problem is illustrated by commit 9991588. There, a serialized object contains a payload that resembles the serialized form of an object, which the serializer using reflection fails to restore correctly. I could imagine a fix for that, but that would imply a change to the format which is *not* backward compatible, so I didn't even start implementing it.

### DIFF
--- a/src/Broadway/Serializer/ReflectionSerializer.php
+++ b/src/Broadway/Serializer/ReflectionSerializer.php
@@ -27,6 +27,10 @@ class ReflectionSerializer implements Serializer
      */
     public function serialize($object): array
     {
+        if (!is_object($object) || $object instanceof \stdClass) {
+            throw new \InvalidArgumentException('Given argument is not serializable.');
+        }
+
         return $this->serializeObjectRecursively($object);
     }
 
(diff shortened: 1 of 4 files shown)


UlrichEckhardt:
Just as a follow-up on this, the types DateTime, DateTimeImmutable and DateTimeZone are also not suitable for serialization, it seems, and those are rather fundamental PHP data types.

othillo:
Serializers are tricky indeed. I've seen companies lose a lot of data due to incorrect serialization.
That's why we recommend to always test your serialization (see this test helper: https://github.com/broadway/broadway/blob/master/src/Broadway/Serializer/Testing/SerializableEventTestCase.php).
If we cannot solve the (edge) cases you encountered with the ReflectionSerializer we could also document this as a known issue together with a purposely failing test.

PR #401, opened 2020-04-12
Replace expected exception annotations with methods

francescotrucchia:


othillo:
Thank you, @francescotrucchia !

PR #402, opened 2020-04-16
[RFC] Replayer

othillo:


src/Broadway/Replaying/Replayer.php
```diff
@@ -0,0 +1,28 @@
...
+
+declare(strict_types=1);
+
+namespace Broadway\Replaying;
+
+use Broadway\EventStore\EventVisitor;
+use Broadway\EventStore\Management\Criteria;
+use Broadway\EventStore\Management\EventStoreManagement;
+
+abstract class Replayer
```
wjzijderveld, line 11: Why is this an abstract class? There's no interface or abstract methods to implement?

wjzijderveld (approved):

PR #404, opened 2020-12-05
added php8 to travis tests

AlessandroMinoccheri:


AlessandroMinoccheri:
Is it time to migrate to Github actions?

AlessandroMinoccheri:
@othillo Finally green CI :)

.travis.yml
```diff
@@ -9,6 +9,7 @@ php:
   - 7.2
```
othillo, line 9: can you remove 7.2 because it is no longer maintained?

examples/event-handling-using-read-models/InvitationProjectorsTest.php
```diff
@@ -28,7 +28,7 @@ protected function createProjector(Broadway\ReadModel\InMemory\InMemoryRepositor
     /**
      * @test
      */
-    public function it_keeps_track_of_the_status_of_an_invitation_when_someone_is_invited()
+    public function itKeepsTrackOfTheStatusOfAnInvitationWhenSomeoneIsInvited()
```
othillo, line 31: can you remove this commit? It has nothing to do with PHP8 and also we like the original casing.

othillo (changes requested):

examples/event-handling-using-read-models/InvitationProjectorsTest.php
AlessandroMinoccheri: Code quality tests don't pass without this commit.

othillo: Hmm, why did they pass before?

AlessandroMinoccheri: I think that with the new versions of PHPCsFixer they have changed the behavior.

AlessandroMinoccheri: @othillo to resolve the problem we can use this configuration for tests to force using snake case

```
'php_unit_method_casing' => ['case' => 'snake_case']
```

What do you think?

othillo: if that would help to remove this commit, I'm +1!

AlessandroMinoccheri: @othillo I have added to the configuration that rule and I have rebased the branch.

src/Broadway/CommandHandling/ClosureCommandHandler.php
```diff
@@ -31,13 +32,22 @@ public function add(\Closure $handler): void
...
+        if ($reflectionType instanceof \ReflectionNamedType) {
+            /** @var \ReflectionNamedType $reflectionNamed */
+            $reflectionNamed = $reflectionParams[0]->getType();
+            if (!$reflectionNamed->isBuiltin()) {
+                $name = new ReflectionClass($reflectionNamed->getName());
+            }
         }
 
-        $index = $reflectionParams[0]->getClass()->getName();
+        if (!isset($reflectionParams[0]) || !$name) {
```
othillo, line 46: no need to check for `!isset($reflectionParams[0]` because on line 35 you already assume it is set.

src/Broadway/CommandHandling/ClosureCommandHandler.php
```diff
@@ -31,13 +32,22 @@ public function add(\Closure $handler): void
         $reflection = new \ReflectionFunction($handler);
         $reflectionParams = $reflection->getParameters();
 
-        if (!isset($reflectionParams[0]) || !$reflectionParams[0]->getClass()) {
```
othillo, line 34: is this needed because of PHP8?

src/Broadway/CommandHandling/ClosureCommandHandler.php
AlessandroMinoccheri: Removed

src/Broadway/CommandHandling/ClosureCommandHandler.php
AlessandroMinoccheri: getClass is not supported by PHP8 so we need to change some parts for deprecation problems .

othillo: Ok, can you add a test for this?

AlessandroMinoccheri: I thought that this class was already covered by `ClosureCommandHandlerTest`, which behavior do you think that needs to have a new test?

othillo: It is tested indeed. I do find this piece of code hard to undertand.

src/Broadway/CommandHandling/ClosureCommandHandler.php
```diff
@@ -31,13 +32,22 @@ public function add(\Closure $handler): void
...
-            throw new ClosureParameterNotAnObjectException();
+        $reflectionType = $reflectionParams[0]->getType();
+
+        $name = null;
+        if ($reflectionType instanceof \ReflectionNamedType) {
+            /** @var \ReflectionNamedType $reflectionNamed */
+            $reflectionNamed = $reflectionParams[0]->getType();
```
othillo, line 40: this is the same as line 35.

src/Broadway/CommandHandling/ClosureCommandHandler.php
```diff
@@ -31,13 +32,22 @@ public function add(\Closure $handler): void
         $reflection = new \ReflectionFunction($handler);
         $reflectionParams = $reflection->getParameters();
 
-        if (!isset($reflectionParams[0]) || !$reflectionParams[0]->getClass()) {
-            throw new ClosureParameterNotAnObjectException();
+        $reflectionType = $reflectionParams[0]->getType();
```
othillo, line 35: Doesn't this fail when there are no parameters? This is what we need a test for.

AlessandroMinoccheri: Ok, thanks for the clarification, I will try to add tests 👍

AlessandroMinoccheri: I Have tried something but it seems that if there are no params it's already covered by `it_throws_when_adding_a_closure_without_an_object_argument` because if there aren't parameters inside `$reflection->getParameters()` there is always something like:

```
array(1) {
  [0]=>
  object(ReflectionParameter)#853 (1) {
    ["name"]=>
    string(7) "command"
  }
}
```

So that test passes because after it doesn't find the correct type and var `$name ` still null. 
So it seems that when there are no parameters it's already covered or I am missing something? 
We can add an if condition to be more defensive that checks if `$reflectionParams` is not empty for example but at the moment it's always filled but the command name for example. 
What do you think?

othillo: Well, there is one parameter in that test: `$params`.

What happens if you add a test like this?

```
/**
     * @test
     */
    public function it_throws_when_adding_a_closure_without_an_object_argument()
    {
        $commandHandler = new ClosureCommandHandler();
        $this->expectException(ClosureParameterNotAnObjectException::class);

        $commandHandler->add(function () { });
    }
```

AlessandroMinoccheri: You are right!
I have added that test and an if condition to check this behavior!

src/Broadway/CommandHandling/ClosureCommandHandler.php
```diff
@@ -31,13 +32,26 @@ public function add(\Closure $handler): void
         $reflection = new \ReflectionFunction($handler);
         $reflectionParams = $reflection->getParameters();
 
-        if (!isset($reflectionParams[0]) || !$reflectionParams[0]->getClass()) {
+        if (empty($reflectionParams)) {
```
othillo, line 35: if (0 === $reflection->getNumberOfParameters())

src/Broadway/CommandHandling/ClosureCommandHandler.php
```diff
@@ -31,13 +32,26 @@ public function add(\Closure $handler): void
...
+            if (!$reflectionNamed->isBuiltin()) {
+                $name = new ReflectionClass($reflectionNamed->getName());
+            }
+        }
+
+        if (!$name) {
+            throw new ClosureParameterNotAnObjectException();
+        }
+
+        $this->handlers[$name->getName()] = $handler;
```
othillo, line 54: can we simplify it to something like this?

```
$reflection = new \ReflectionFunction($handler);
if (0 === $reflection->getNumberOfParameters()) {
    throw new ClosureParameterNotAnObjectException();
}

$reflectionType = $reflection->getParameters()[0]->getType();
if ($reflectionType instanceof \ReflectionNamedType && !$reflectionType->isBuiltin()) {
    $this->handlers[$reflectionType->getName()] = $handler;
} else {
    throw new ClosureParameterNotAnObjectException();
}
```

AlessandroMinoccheri: I was thinking about this solution, usually, I don't like to use `else`  condition but sometimes creates a more clear code. We can think to add it now or trying to avoid  `else` but changing with something like this:

```
        if (!$reflectionType instanceof \ReflectionNamedType || $reflectionNamed->isBuiltin()) {
            throw new ClosureParameterNotAnObjectException();
        }

        $name = new ReflectionClass($reflectionParams[0]->getType()->getName());
```
Again, I don't like to use negative conditions, it's always a trade-off between best practice and readable code. And usually, I prefer readable code, but it's another solution.

In my PR I have splitter `isBuiltIn` check I think for some check failed

src/Broadway/CommandHandling/ClosureCommandHandler.php
```diff
@@ -31,13 +32,26 @@ public function add(\Closure $handler): void
...
         }
 
-        $index = $reflectionParams[0]->getClass()->getName();
+        $reflectionType = $reflectionParams[0]->getType();
 
-        $this->handlers[$index] = $handler;
+        $name = null;
+        if ($reflectionType instanceof \ReflectionNamedType) {
+            /** @var \ReflectionNamedType $reflectionNamed */
+            $reflectionNamed = $reflectionParams[0]->getType();
```
othillo, line 44: This is duplicate code and still not very readable. Can you simplify it based on my earlier suggestion?

othillo (changes requested):

src/Broadway/CommandHandling/ClosureCommandHandler.php
AlessandroMinoccheri: Ok, sure!

src/Broadway/CommandHandling/ClosureCommandHandler.php
```diff
@@ -15,6 +15,7 @@
 
 use Broadway\CommandHandling\Exception\ClosureParameterNotAnObjectException;
 use Broadway\CommandHandling\Exception\CommandNotAnObjectException;
+use ReflectionClass;
```
othillo, line 18: This is not used.

othillo (approved):

othillo:
Thank you, @AlessandroMinoccheri !

AlessandroMinoccheri:
Thanks, @othillo for review this PR.

PR #414, opened 2021-09-03
added docker, updated namespaces, updated dependencies

auselin:
- added docker
- updated namespaces
- updated dependencies


wjzijderveld:
Was this intented to be a PR here? Because I'm very confused why we would want to merge this?

PR #420, opened 2022-04-22
update supported PHP versions

othillo:


AlessandroMinoccheri:
Hi @othillo by removing the supported version do you mean also to update the `composer.json` file with an higher supported version of PHP?

wjzijderveld (approved):

wjzijderveld:
It's a bit of a tricky thing @AlessandroMinoccheri  There's no reason to assume it won't work on 7.2, we just don't test against it as that PHP version itsn't supported anymore.
To avoid confusion it's probably best to update composer.json as well, even if it means some people can't update anymore even though it would work fine.

PR #423, opened 2022-09-05
[RFC] Upcasting

matiux:
With this PR I tried to develop my idea up upcasting for the broadway library. It is based on a previous PR, but I have also provided an example of how to use this functionality via a test.

If the idea is approved, I would like to then work on the DBAL driver and symphony bundle for broadway.

**How it works:**

At the moment, the code works with the InMemoryEventStore class. In order to not touch the original class, I created a decorator for it. Regarding the DBAL driver integration, I could apply the decorator pattern or modify the original DBALEventStore class. What do you think?

The SequentialUpcasterChain class is a kind of registry used to register individual upcasters, which obviously need to be registered in an orderly manner.

I could potentially improve the SequentialUpcasterChain class by handling the upcasters' priority during instantiation in order to guarantee and enforce the ordering.

The example I have provided is based on the UserCreated event, which obviously is a bit of a stretch as it is conceptually an event that would be dispatched only once for each aggregate, but it should give the idea of how the feature works. A more realistic use case could be based on an event like PersonalDataRegistered, which could occur several times.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -2,3 +2,5 @@
 composer.lock
 .php-cs-fixer.cache
 .phpunit.result.cache
+.idea
+docker
--- a/.gitignore
+++ b/.gitignore
@@ -2,5 +2,3 @@
 composer.lock
 .php-cs-fixer.cache
 .phpunit.result.cache
-.idea
-docker
--- a/src/Broadway/Upcasting/UpcastingInMemoryEventStore.php
+++ b/src/Broadway/Upcasting/UpcastingInMemoryEventStore.php
@@ -1,0 +1,80 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Broadway\Upcasting;
+
+use Broadway\Domain\DomainEventStream;
+use Broadway\Domain\DomainMessage;
+use Broadway\EventStore\EventStore;
+use Broadway\EventStore\EventVisitor;
+use Broadway\EventStore\InMemoryEventStore;
+use Broadway\EventStore\Management\Criteria;
+use Broadway\EventStore\Management\EventStoreManagement;
+
+class UpcastingInMemoryEventStore implements EventStore, EventStoreManagement
+{
+    /**
+     * @var InMemoryEventStore
+     */
+    private $eventStore;
+
+    /**
+     * @var UpcasterChain
+     */
+    private $upcasterChain;
+
+    public function __construct(InMemoryEventStore $eventStore, UpcasterChain $upcasterChain)
+    {
+        $this->eventStore = $eventStore;
+        $this->upcasterChain = $upcasterChain;
+    }
+
+    public function load($id): DomainEventStream
+    {
+       return $this->upcastStream(
+           $this->eventStore->load($id),
+           $id
+       );
+    }
+
+    public function loadFromPlayhead($id, int $playhead): DomainEventStream
+    {
+        return $this->upcastStream(
+            $this->eventStore->loadFromPlayhead($id, $playhead),
+            $id
+        );
+    }
+
+    private function upcastStream(DomainEventStream $eventStream, $id): DomainEventStream
+    {
+        $upcastedEvents = [];
+
+        foreach ($eventStream as $domainMessage) {
+            $upcastedEvents[] = $this->createUpcastedDomainMessage($domainMessage, $id);
+        }
+
+        return new DomainEventStream($upcastedEvents);
+    }
+
+    private function createUpcastedDomainMessage(DomainMessage $domainMessage, $id): DomainMessage
+    {
+        return new DomainMessage(
+            $id,
+            $domainMessage->getPlayhead(),
+            $domainMessage->getMetadata(),
+            $this->upcasterChain->upcast($domainMessage->getPayload()),
+            $domainMessage->getRecordedOn()
+        );
+    }
+
+    public function append($id, DomainEventStream $eventStream): void
+    {
+        $this->eventStore->append($id, $eventStream);
+    }
+
+    public function visitEvents(Criteria $criteria, EventVisitor $eventVisitor): void
+    {
+        $this->eventStore->visitEvents($criteria, $eventVisitor);
+    }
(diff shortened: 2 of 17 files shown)


matiux:
@othillo do you have 5 minutes to take a look?

matiux:
@wjzijderveld @othillo I don’t want to be insistent, simply put, I would just like to know how come my PR didn’t get any feedback. Is it because such functionality shouldn’t be part of the broadway library? Is there something wrong with the implementation? If the answer is the latter, could you provide me with some feedback, and I will adapt the code.

src/Broadway/Upcasting/UpcastingInMemoryEventStore.php
```diff
@@ -0,0 +1,80 @@
...
+     * @var InMemoryEventStore
+     */
+    private $eventStore;
+
+    /**
+     * @var UpcasterChain
+     */
+    private $upcasterChain;
+
+    public function __construct(InMemoryEventStore $eventStore, UpcasterChain $upcasterChain)
```
wjzijderveld, line 27: I like how you changed it to be a wrapper instead of adjusting existing EventStoers, but ideally we'd just have an `UpcastingMemoryStore` that wraps an `EventStore`, is there a specific reason you made this specific for the InMemory implementation here?

wjzijderveld (commented):
Sorry for the silence, it's been difficult to focus on OSS (and just as hard to make time for it). So especially a subject like upcasting is hard to give the proper attention in that case.

Have you used upcasting in production by any chance?

matiux:
> Sorry for the silence, it's been difficult to focus on OSS (and just as hard to make time for it). So especially a subject like upcasting is hard to give the proper attention in that case.
> 
> Have you used upcasting in production by any chance?

No problem, I'm glad that you have reopen the PR! By the way, soon I will need to use code that I have push in the PR, in a production project. This make me able to work and improve this feature continuing with contributes.

src/Broadway/Upcasting/UpcastingInMemoryEventStore.php
matiux: For now I have preferred decorate original `InMemoryEventStore` not to make my feature "invasive". However I'm agre with you, in this way I should decorate also `EventStore` (DBAL). Another convenient way could be to move the upcasting behaviour in the `EventSourcingRepository`. By decoration or adding the `UpcasterChain` in the constructor. After all `UpcasterChain` is only a registry, and if it is empty, will do not nothing

wjzijderveld: I think you misunderstood, I don't think there's any need for a storage specific UpcasterEventStore.

```php

final class UpcastingEventStore implements EventStore, EventStoreManagement
{
  public function __construct(EventStore $eventStore, UpcasterChain $upcasterChain) {}

  // [..] the rest of the methods just like you have them.
}
```

.gitignore
```diff
@@ -1,4 +1,6 @@
 /vendor/
 composer.lock
 .php-cs-fixer.cache
-.phpunit.result.cache
```
wjzijderveld, line 4: Can you revert this btw? No need to have those specifics in this project, otherwise we'll end up with an endlessly long gitignore file with all kinds of non-project related stuff :)

matiux:
I should have applied your tips and it is actually cleaner that way

matiux:
Hi @wjzijderveld what do you think about last commits?

wjzijderveld (commented):
It does look better, although it's a bit sad we can't typehint the $eventStore argument now. But as PHP7.4 is EOL anyway, maybe we need to start requiring PHP8.0 and then we would be able to use the union type. 

Second thing that I'm not sure about (somehow I didn't see that earlier), is if the upcaster should have access to the whole DomainMessage instead of just the payload. Especially the recordedOn could be quite useful in some cases I think. 

@othillo what do you think?

matiux:
I mean what you say. And yes, it might make sense to pass the entire DomainMessage. Let's hear what @othillo says and if necessary I can make the change

Regarding the switch to php 8, I'd be happy to do a second PR

matiux:
@othillo @wjzijderveld  is there any news? :)

matiux:
Hi @wjzijderveld , @othillo ,

How are things going? I just got back to work after my Christmas holidays but noticed that this PR is still open. Are there any changes you’d prefer me to make so this PR can be approved?

 
Thanks

matiux:
Hi @othillo I have seen activities on the PR front. Can you also take a look at the dialogue that took place between me and @wjzijderveld ?
Thanks

othillo:
> Second thing that I'm not sure about (somehow I didn't see that earlier), is if the upcaster should have access to the whole DomainMessage instead of just the payload. Especially the recordedOn could be quite useful in some cases I think.
> 
> @othillo what do you think?

I agree, I think it would be more useful to upcast a DomainMessage and not just its payload.

othillo:
> Hi @othillo I have seen activities on the PR front. Can you also take a look at the dialogue that took place between me and @wjzijderveld ? Thanks

Hey @matiux, thank you for contributing! Its almost good to go but I would like the upcaster to upcast a DomainMessage instead of just its payload. Would you mind adding tests for the classes in `/src` ?

matiux:
Hi @othillo thanks for your reply!
Sure, I will work on `DomainMessage` upcasting instead of upcasting only the payload and I will let you know asap. I will also update the example and add some tests.

Thank you!!

matiux:
Hi @othillo , I think to have done all changes you have requested to me. Now the subject of the upcasting is the DomainMessage and no longer payload. Moreover I have added 2 unit test. 
In the examples folder you can find more tests to see three real use cases.
What do you think about this last commit?

Cc @wjzijderveld

wjzijderveld:
Thanks for your contributions and your patience @matiux
Sorry it took so long!

matiux:
No problem @wjzijderveld, thanks!
When could we have a tag with this commit?

PR #458, opened 2025-08-05
Feature | Update to PHP 8.4

Bakhtarian:
* Update composer to accept PHP 8.4
* Update PHPUnit to newest version
* Rework tests according to newest PHPUnit version


Cosmologist:
Looks like Broadway is dead - I made a mistake choosing it as the library for event-sourcing

Bakhtarian:
> Looks like Broadway is dead - I made a mistake choosing it as the library for event-sourcing

Perhaps it's best to setup your own implementation of Event Sourcing. I also created one for learning purpose, but the ground-works are pretty straight forward. You can look at my [own implementation](https://github.com/Bakhtarian/todo_list) on how I approached it. Setting it up in Symfony is not that complicated when you use the Symfony messenger component. I do understand that it's not ideal to set it up yourself, and that these kind of repositories come in handy when setting up your ES project. If you really want to have a ready-to-use project, you can always look at Prooph or EventSauce. I know that EventSauce is definitely still maintained. Prooph also had some components deprecated. The goal of my PR was to have the setup ready for PHP 8+ projects. Including the tests + correct type hinting for the IDE. Unfortunately it does indeed seem like it won't be included.

